PR #4, opened 2016-03-21
[Bug Fix] Apply width on cell component based on it's columns width

steffenbrem:
Right now when you leave out `{{t.head}}`, the table does not keep it's width per column, since they are only defined on the `thead`. With those changes it will always have the correct width.

### DIFF
--- a/addon/components/lt-cell.js
+++ b/addon/components/lt-cell.js
@@ -10,6 +10,7 @@ export default Ember.Component.extend({
   layout,
   tagName: 'td',
   classNames: ['lt-cell'],
+  attributeBindings: ['width'],
   classNameBindings: ['align', 'isSorted'],
 
   column: null,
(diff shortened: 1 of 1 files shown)


addon/components/lt-cell.js
```diff
@@ -22,6 +23,8 @@ export default Ember.Component.extend({
 
   isSorted: computed.oneWay('column.sorted'),
 
+  width: computed.alias('column.width'),
```
steffenbrem: I've sorted the computed properties alphabetically.

offirgolan: Can you change that, as well as isSorted to computed.readOnly?

steffenbrem: Sure! :+1:

steffenbrem:
I've squashed it and applied your feedback.

offirgolan:
:+1: Thank you! Ill merge this as soon as I figure out why all my ember-beta builds keep failing.

PR #30, opened 2016-06-03
Support cell and column types

offirgolan:


offirgolan:
ping @thoov. What do you think about this feature? 

It allows us and users to create cell and column 'types' which should extend from the base. This allows for the possibility of creating draggable columns, link-to cells, etc.

addon/templates/components/lt-row.hbs
```diff
@@ -1,3 +1,7 @@
 {{#each columns as |column|}}
-  {{lt-cell column=column row=row rawValue=(get row column.valuePath) tableActions=tableActions}}
+  {{component (concat 'light-table/cells/' column.cellType)
+    column=column
```
thoov: This is a minor thing but instead of `column=column` and `row=row` you might want to look into [positionalParams](http://emberjs.com/api/classes/Ember.Component.html#property_positionalParams)

thoov:
@offirgolan I think this looks good! Do you have an example written up yet (aka on your demo page)?

PR #32, opened 2016-06-06
Use Trackpad Scroll Emulator for better scrollbars

taras:
I ended up implementing this because it would take a lot of work to explain.

My primary objective was to eliminate native scrollbar from table body. The native scrollbar is quite ugly and it messes with column positioning on Windows. 

I also ended up addressing a problem with table heights. Previously table heights were set on `t.body` which was weird to style because you needed to manually consider the height of footer and header. The improved implementation uses Flexbox to automatically calculate table body taking into consideration header and footer.

I removed table-scroll mixin in favour of scrolling mechanism provided by Trackpad Scroll Emulator. I don't think TSE's mechanism is better but we don't need two versions of the same functionality.

I added several dependencies:
- ember-composable-helper - provides pipe and compute helpers
- ember-scrollable - provides TSE and `as-scrollable` component


addon/components/lt-body.js
```diff
@@ -92,6 +91,15 @@ export default Ember.Component.extend({
   multiSelect: false,
 
   /**
+   * Hide scrollbar when not scrolling
+   *
+   * @property autoHide
+   * @type {Boolean}
+   * @default true
+   */
+  autoHide: true,
```
offirgolan: Maybe we should rename this to `autoHideScrollbar` to give more context?

addon/components/lt-body.js
```diff
@@ -198,6 +199,15 @@ export default Ember.Component.extend({
      */
     onRowDoubleClick( /* row */ ) {
       callAction(this, 'onRowDoubleClick', ...arguments);
+    },
+    
+    /**
+     * onScrollEnd action.
+     * 
+     * @event onScrollEnd
```
offirgolan: `onScrollEnd` --> `onScrolledToBottom`

addon/components/lt-body.js
taras: Yes, will do this now.

package.json
```diff
@@ -42,6 +41,7 @@
     "ember-cli-uglify": "^1.2.0",
     "ember-cli-yuidoc": "0.8.3",
     "ember-code-snippet": "^1.2.1",
+    "ember-composable-helpers": "0.26.0",
```
offirgolan: I believe this needs to be added as a dependency

taras: Done

addon/components/lt-body.js
taras: Done

addon/components/lt-body.js
taras: Done

addon/templates/components/lt-body.hbs
```diff
@@ -1,8 +1,14 @@
...
+{{#lt-scrollable 
+  autoHide=autoHideScrollbar
+  scrollEnd="onScrolledToBottom"
+  scrollBuffer=scrollBuffer
+  as |scrollbar|
+}}
 <table>
   <tbody class="lt-body">
-    {{#each rows as |row|}}
+    {{#each (compute (pipe scrollbar.update) rows) as |row|}}
```
offirgolan: This is pretty confusing and will be for anyone that attempts to read this. Can you add a brief comment?

taras: Removed pipe and added comment.

offirgolan:
Awesome work @taras! Thank you! 😺

PR #33, opened 2016-06-07
Table height fixes

taras:
- Added test that checks that table consumes 100% height of the container
- Fixed problem where body was not consuming all of the space when it had only 1 row (test not working in Phantomjs - commented it out)


offirgolan:
Thank you 😺

PR #42, opened 2016-06-09
Add hideable option to columns

### LINKED ISSUE #41: Add a "hideable" option to configure if column can be hidden
The `Column` class has the `sortable` option to configure if it can be sorted or not. It would be nice to also have an option to configure if the column can be hidden or not.

Having this option in the `Column` class will be useful to create a generic component for the user to choose the columns visible in the table, something like a list with a checkbox for each column in the table that can be hidden.

offirgolan:
Resolved #41


taras:
Why not just change the columns when you want to hide a column?

offirgolan:
@taras I think this is more of a convenience thing. That way you can easily query which columns can be hidden and create a toggle menu sort of thing. This also adds the `is-hideable` class to easily target and theme such columns.

taras:
hmm... ok :)

dspigarelli:
Changing the columns as @taras suggests would meant targeting the exact column and removing with `removeColumn*` to avoid losing `sorted` status on columns that aren't changing. I find the `hidden` flag much easier to use/consume.

PR #50, opened 2016-06-12
Fixed full height demo

taras:
- Switched to release version of ember-scrollable
- Removed table container
- Made full-container a full container to force table height


offirgolan:
Thanks Taras 😄

PR #51, opened 2016-06-13
added draggable column type and dummy example

HenryVonfire:


addon/components/columns/draggable.js
```diff
@@ -0,0 +1,77 @@
...
+  classNameBindings: [
+    'align',
+    'isGroupColumn:lt-group-column',
+    'isHideable',
+    'isSortable',
+    'isSorted',
+    'column.classNames',
+    'droptargetClass',
+    'dragClass'
+  ],
```
offirgolan: You dont need to include all class binding from the Base class since this property get concatenated.

addon/components/columns/draggable.js
```diff
@@ -0,0 +1,77 @@
...
+      });
+
+      let columns = table.columns;
+      let temp = columns.objectAt(draggedIndex);
+      columns.replace(draggedIndex, 1, [columns.objectAt(staticIndex)]);
+      columns.replace(staticIndex, 1, temp);
+    }
+  },
+
+  droptargetClass: computed(function () {
```
offirgolan: this can be replaced with `classNames: ['droptarget']`

addon/components/columns/draggable.js
```diff
@@ -0,0 +1,77 @@
...
+  }),
+
+  dragLeave(event) {
+    event.preventDefault();
+    this.set('dragClass', null);
+  },
+
+  dragOver(event) {
+    event.preventDefault();
+    this.set('dragClass', 'dragged-over');
```
offirgolan: This can be replaces by adding `isDragging` to the classNameBindings, then setting `isDragging` to true/false in dragOver / dragLeave

offirgolan:
@HenryVonfire, I made a few comments. Im assuming this still doesnt work for grouped columns... Let me know what you need from me to make that happen 😺

addon/components/columns/draggable.js
```diff
@@ -0,0 +1,77 @@
...
+    let staticName = this.get('column.valuePath');
+    if (draggedName !== staticName) {
+      const table = this.get('table');
+      let draggedIndex, staticIndex, draggedColumn, staticColumn;
+      table.columns.forEach((column, index) => {
+        if (column.valuePath === draggedName) {
+          draggedIndex = index;
+          draggedColumn = column;
+        } else {
+          if (column.valuePath === staticName) {
```
offirgolan: concat this to `else if()`

HenryVonfire:
I've added the swapping functionality between groups; now when starting the dragging it recognises if the dragged column is a parentcolumn or a subcolumn and when it's dropped, only allows to swap subcolumns of the same group.

There's still a problem in which the content of the table isn't updated when swapping subcolumns.

taras:
@HenryVonfire thank you for your contribution and for taking the time to integrate @offirgolan's feedback. @offirgolan and I reviewed this PR and I have a few comments that effect the next steps with this PR.

From my experience, being able to reorder columns is a very common table feature. It's often useful to be able to reorder columns of different types. If a column is draggable then it can't be of another type. This is too limiting.

I would like to propose that you and I work together to refactor this PR to allow columns to be draggable without using type. I'd like to experiment with using something like [ember-sortable](https://github.com/jgwhite/ember-sortable) to allow tables to be sorted. 

What do you think?

HenryVonfire:
I didn't know about `ember-sortable` and HTML5 drag system was pretty easy to implement so that's why I chose it :sweat_smile:

Of course I'm glad to work with you, I'm sure that I'll learn a lot! :smile: what do you have in mind? do you need me to give you access to the repository?

eddiesholl:
In case its useful, I've got an attempt at this functionality at https://github.com/eddiesholl/ember-light-table/tree/reorder-columns. It's mostly complete, just no group support.

My approach requires more code, as I'm using mouse events instead of just drag. One path I went down that might be useful is managing the drag and drop in table-header.js. It might help with things like applying the order change back into the table?

HenryVonfire:
Closed in favor of this new PR [109](https://github.com/offirgolan/ember-light-table/pull/109)

PR #55, opened 2016-06-13
Support templates in column and cell types

### LINKED ISSUE #53: Support templates in Cell & Column types

offirgolan:
Resolved #53 

Running `ember g cell-type link-to`

will now create a new template allowing you to do something like this

```hbs
{{#link-to column.route row.id}}
  {{yield this}}
{{/link-to}}
```

@taras I really like this, but I think the implementation is a little raw. Would love some feedback! I dont really know how to avoid those partials as well as make testing a bit easier....


offirgolan:
After discussion with @taras. We will be putting this future on hold.

PR #60, opened 2016-06-15
Recalculate scrollbar when viewport dimensions change

offirgolan:
When a dom element is removed or inserted, we dont recalculate the scrollable container. @taras this solution works but I believe can be improved.

### DIFF
--- a/addon/components/lt-scrollable.js
+++ b/addon/components/lt-scrollable.js
@@ -1,5 +1,31 @@
 import ScrollableComponent from 'ember-scrollable/components/scrollable';
 
 export default ScrollableComponent.extend({
-  classNames: ['lt-scrollable']
+  classNames: ['lt-scrollable'],
+
+  didInsertElement() {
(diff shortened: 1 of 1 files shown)


taras:
@offirgolan which DOM element is removed or inserted?

offirgolan:
@taras lets say you have a flexbox container with flex-direction: row. You initially have a single item in that container (i.e. the table) then some event gets triggered and a new item is added to the container. Now, the table's width has been reduced by 50% but the scrollable container does not trigger its recalculate method due to the fact that a resize event hasnt been triggered

taras:
oh, I see. I think the solution that I linked to ember-scrollable should do it. It adds a brief animation to the height of the element. When the height changes, the animations' callback is triggered which triggers an event similar to resize event. 

That's likely the better solution to this problem.

offirgolan:
Perfect. Lets prioritize that since I believe this is currently a blocker (especially in my current apps)

PR #76, opened 2016-06-24
Introduce virtualScrollbar and use ember-in-viewport for infinite scroll handling

taras:
# Changes
- Added ember-element-resize-detector to monitor the size of light-table component. When size changes scrollbars will update. This only relevant and active in virtualScrollbar=true mode
- Refactored to make ember-scrollable optional. It'll only be activated when `light-table` is rendered with `virtualScrollbar=true`
- Made it possible to set `virtualScrollbar` on the table or the body
- Updated examples to use `virtualScrollbar=true`


addon/templates/components/lt-body.hbs
```diff
@@ -12,7 +13,7 @@
         {{yield columns rows}}
       {{else}}
         {{! scrollbar.update action when rows changes and evaluates to rows !}}
```
offirgolan: This comment is no longer needed

taras: Yep, I should remove. I'm not sure if change is 100% safe, but if it's not I'll address it.

addon/styles/addon.css
```diff
@@ -1,5 +1,5 @@
 .ember-light-table {
-  height: auto;
+  height: inherit;
```
offirgolan: Setting this to inherit, allow the table to get its height from the parent container

PR #79, opened 2016-06-25
Added helper that can be used for simple tables

taras:
# TODO
- [x] Created table helper
- [x] Added simple unit test for test helper
- [x] Added integration test to show helper working
- [] Documentation

### DIFF
--- a/addon/helpers/table.js
+++ b/addon/helpers/table.js
@@ -1,0 +1,8 @@
+import Ember from 'ember';
+import Table from '../classes/Table';
+
+export function table([columns, data = []]) {
+  return new Table(columns, data);
+}
+
+export default Ember.Helper.helper(table);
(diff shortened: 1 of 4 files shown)


offirgolan:
@taras I think we can released this now that enableSync has been released.

```hbs
{{light-table (table columns (get data '[]') enableSync=true)}}
```

buschtoens:
:bellhop_bell: :grin:

offirgolan:
I think this is a pretty dangerous route since ever time the helper will recompute, it will create a new table instance which will reset the table's entire state. That can be pretty hellish to debug....

## Proposed Change

```hbs
{{light-table (table columns (get data '[]') enableSync=true)}}
```

```js
export default Ember.Helper.extend({
  table: null,

  compute([columns = [], data = []), hash) {
    let table = this.get('table');
    
    if (!table) { 
      table = new Table(columns, data, hash);
      this.set('table', table);
      return table;
    }
    
    // We still lose the column's state here
    // Maybe we can do a diff here to see what changed? 
    table.setColumns(columns);
    table.setRows(data);

    return table;
  }
});
```

PR #91, opened 2016-07-01
Add some cellType and columnType docs

rwwagner90:
I did some destructuring as well, for consistency of not using `Ember.`, but destructuring instead.


taras:
💯 thank you!

rwwagner90:
No problem @taras! Trying to help out where I can 😄

offirgolan:
@rwwagner90 thanks for taking the time to document the cell and column types, I think it looks great! In terms of styling, Im not really a fan of declaring `Component` as a constant (especially if there is no other constants in the rest of the file).

rwwagner90:
@offirgolan I just try to destructure whatever I can. Is there a reason you particularly do not like doing so with Component?

offirgolan:
Honestly, its nothing more than of a personal preference / pet-peeve

rwwagner90:
@offirgolan I'm just not sure what makes component special. I thought the standards were you should destructure everything out of the Ember namespace that you can rather than using `Ember.Component` `Ember.Route` etc. I'll be glad to change it back if you'd like though.

offirgolan:
Thanks @rwwagner90

PR #94, opened 2016-07-01
Add options hash to row creation methods

### LINKED ISSUE #93: Support options hash when invoking row-related methods like Table.addRow()
Currently, when invoking row-related methods like `Table.addRow()`, the options hash doesn't make it to the Row constructor:

```
this.set('table', new Table(columns = [], rows = []));

this.get('table').addRow({
  key: 'value'
}, {
  classNames: 'some-class-name'
});

// class name is not set on the row
```

To get the options to be set on a row, you currently have to initialize a Row first.

```
let row = new Row({
  key: 'value'
}, {
  classNames: 'some-class-name'
});

this.get('table').addRow(row);
```

Would it not be great if you could pass along the options along with the data when using methods like addRow or pushRow?

offirgolan:
Resolves #93

### DIFF
--- a/addon/classes/Table.js
+++ b/addon/classes/Table.js
@@ -139,22 +139,24 @@ export default class Table extends Ember.Object.extend({
    * Replace all the row's content with content of the argument. If argument is an empty array rows will be cleared.
    * @method setRows
    * @param  {Array} rows
+   * @param  {Object} options
    * @return {Array} rows
    */
-  setRows(rows = []) {
-    return this.rows.setObjects(Table.createRows(rows));
+  setRows(rows = [], options = {}) {
+    return this.rows.setObjects(Table.createRows(rows, options));
   }
 
   /**
    * Push the object onto the end of the row array if it is not already present.
    * @method addRow
    * @param  {Object} row
+   * @param  {Object} options
    */
-  addRow(row) {
+  addRow(row, options = {}) {
     if (row instanceof Row) {
       this.rows.addObject(row);
     } else if (isNone(this.rows.findBy('content', row))) {
-      this.pushRow(row);
+      this.pushRow(row, options);
     }
   }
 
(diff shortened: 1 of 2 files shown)


addon/classes/Table.js
```diff
@@ -139,42 +139,46 @@ export default class Table extends Ember.Object.extend({
...
 
   /**
    * Push the objects onto the end of the row array if it is not already present.
    * @method addRows
    * @param  {Array} rows
+   * @param  {Object} options
    */
-  addRows(rows = []) {
-    rows.forEach(r => this.addRow(r));
+  addRows(rows = [], options = {}) {
```
curtisgr: @offirgolan not sure this is correct.

offirgolan: @curtisgr why's that? `options` passed to `addRows` will be applied to all row objects.

curtisgr: @offirgolan so, for example, I'm iterating over a collection and depending on certain states, I'm decorating each row with a set of class names.  In my case, I would not want the same options hash applied to each row.

curtisgr: For example:

```
// other stuff

let rows = [];

while (date.format('YYYY-MM-DD') !== endDate.format('YYYY-MM-DD')) {
  let classNames = [];

  if (business.isWeekendDay(date)) {
    classNames.pushObject('is-weekend-day');
  }

  if (business.isFederalHoliday(date)) {
    classNames.pushObject('is-federal-holiday');
  }

  let holiday = business.federalHolidays.findBy('date', date.format('YYYY-MM-DD'));

  let row = new Row({
    date: date.format('YYYY-MM-DD'),
    notes: holiday ? holiday.name : ''
  }, {
    classNames: classNames
  });

  rows.pushObject(row);

  date = date.add(1, 'days');
}

this.get('table').addRows(rows);
```

offirgolan: @curtisgr I totally understand your use-case. You have 2 options. 

1. In your while loop, just do 

```js
this.table.addRow({
    date: date.format('YYYY-MM-DD'),
    notes: holiday ? holiday.name : ''
  }, {
    classNames: classNames
  });
```

2. Instead of using `new Row` use the static method `Table.createRow{{}, {})` which is public API. 

In conclusion, I think either solution would work and will be using public APIs. Do you have some other proposition?

curtisgr: works for me!  thanks!

offirgolan: 👍

PR #101, opened 2016-07-13
Always use Ember object `get`.

blimmer:
In many places, we're currently accessing with this.varName, which prevents people using computed properties.

Also, in my experience, this can be a confusing anti-pattern for those new to Ember. Many will copy from the demo app, so want to make sure people get used to this pattern.


offirgolan:
Thanks 😸!

offirgolan:
@blimmer I went ahead and published a patch release

PR #102, opened 2016-07-13
Use ember-font-awesome vs. ember-cli-font-awesome.

blimmer:
Subtle distinction, but since this is an Ember >2.3 app, you should use the flavor without the `cli`. This squashes those pesky build-time warnings about overriding `init` without `_super`.


offirgolan:
Ooooo okay, I didnt know there an `ember-font-awesome`. Thanks @blimmer

PR #104, opened 2016-07-14
Make text unselectable if column is sortable

blimmer:
I find myself accidentally highlighting the table headers when clicking to sort:

### DIFF
--- a/addon/styles/addon.css
+++ b/addon/styles/addon.css
@@ -61,6 +61,7 @@
 
 .ember-light-table .lt-column.is-sortable {
   cursor: pointer;
+  user-select: none;
 }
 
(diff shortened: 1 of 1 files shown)


blimmer:
not sure why that one test failed - looks like bower might have blipped.

offirgolan:
I dont know if I want to force unelectable text on the table columns. I feel like that can be easily set up by the user. At the same time, it can be argued that we can include this and it easy for the user to revert...

@taras thoughts on this?

taras:
I don't have a strong opinion about this and I agree with your assessment. Worst case scenario, the user can easily overwrite the CSS.

offirgolan:
👍

PR #109, opened 2016-07-21
[FEATURE] Draggable Columns

offirgolan:
@HenryVonfire i've decided to tackle this on! I got mostly everything working (a bit of code taken from your PR) but it feels kinda laggy / slow. Plus, the column width change while dragging a column over.... if you can help me out with that, we can ship this feature out soon!

@taras have you had any experience with ember-sortable? It feels a bit off tbh. Maybe it might be more optimal to do it via dnd-helpers... 

- [x] Work with simple headers 
- [x] Work with grouped headers
- [ ] Work with footer
- [ ] Might need to rework how column groups and sub columns are rendered to get footer logic working 
- [ ] Fix laggy feel when dragging a column


vasilakisfil:
Looking forward to it!

HenryVonfire:
As far as I know, it's like it adds the width of a column when you start to drag a column (i.e. if you have 3 columns it places the first and the second correctly but the third in the place of a fourth column when you drag one). I was looking at the `update` method in the `sortable-group` component but I haven't found the problem yet.

I also removed the CSS of ember-light-table to be sure it was because of it but the problem still persists.

PR #122, opened 2016-07-26
Remove deprecations + Some cleanup

### LINKED ISSUE #100: deprecation warning in tests
```bash
DEPRECATION: You modified (or sharedOptions.fixedHeader sharedOptions.fixedFooter) twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 3.0 [deprecation id: ember-views.render-double-modify]
        at logDeprecationStackTrace (http://localhost:7357/assets/vendor.js:16449:19)
        at HANDLERS.(anonymous function) (http://localhost:7357/assets/vendor.js:16556:7)
        at raiseOnDeprecation (http://localhost:7357/assets/vendor.js:16479:12)
        at HANDLERS.(anonymous function) (http://localhost:7357/assets/vendor.js:16556:7)
        at invoke (http://localhost:7357/assets/vendor.js:16572:7)
        at deprecate (http://localhost:7357/assets/vendor.js:16540:32)
        at Object.deprecate (http://localhost:7357/assets/vendor.js:25882:37)
        at Class.exports.default._emberMetalMixin.Mixin.create._Mixin$create.scheduleRevalidate (http://localhost:7357/assets/vendor.js:52825:26)
        at http://localhost:7357/assets/vendor.js:23046:32
```

offirgolan:
Resolves #100 

- Renamed `lt-column-resize` to `lt-column-resizer`
- Fixed deprecation warning caused by checking fixedHeader & fixedFooter twice on the same render
- Fixed deprecation caused by multiple renders in lt-infinity by forcing each observer call to execute on the `afterRender` queue of the run loop


taras:
It's unfortunate that we need to do all of this hackary with the runloop to get this to work properly. Hopefully we'll find another approach in the future.

offirgolan:
@taras I agree but I think the current implementation is rather clean and self explanatory. Maybe in the future we can think of a better way to do the whole `sharedOptions` stuff

Issue #114, opened 2016-07-22
Fixed left column


bakerac4:
One of the requirements I was given was to implement the first column as a fixed column. Much like http://opensource.addepar.com/ember-table/#/overview. If you collapse the width of the page, you will see that you can scroll everything but the first column to the right. This gives a much better presentation on mobile/tablet. Can you guys think of a way that I might accomplish this?

taras:
@bakerac4 we don't have horizontal scroll in this addon.

I'm working on a few large tables right now and we're implementing a different kind of responsive behaviour. Instead allowing horizontal scrolling, we're collapsing hidden columns into an accordion with rows. This feature will likely be quiet portable and we could add it to this table. 

Speaking for myself, I don't see horizontal scrolling as a feature that we should be investing into. It complicates the implementation considerably and having horizontal and vertical scroll is generally not a practical user experience.

If you absolutely have to build this, then I could give you some suggestions but I'd advice exploring alternative user experiences.

bakerac4:
Hey @taras do you have an example I can view for the hidden/collapsable column tables your working? I would love to be able to play around with it to see how it feels? It sounds like an interesting idea.

taras:
@bakerac4 I don't have one yet, but I'll have a finished implementation in about 1 month.

PR #131, opened 2016-07-30
[WIP] Fixed columns

cball:
This has a decent amount more for me to do before I'd consider it mergeable but I wanted to open it up for discussion.

This library is gaining traction, so I wanted to look at adding some of the high level features developed in https://github.com/cball/justa-table so that I can put resources towards this library instead. We currently have 3 projects that need fixed column support and I know others that have expressed the need for that feature as well. @taras had concerns using fixed columns in https://github.com/offirgolan/ember-light-table/issues/114. While complicated to implement, it is a pretty common ask for complex tables and I don't think you can always just implement a different UI.

First, the *largest issue*: Trackpad Scroll Emulator which is used by `ember-scrollable` can only scroll 1 way. (see https://github.com/jnicol/trackpad-scroll-emulator#5-limitations). As such, in this PR when there are fixed columns present, native scrollbars are used instead. I'd actually argue to put the work in to make native scrollbars work across the board but that's a separate issue.

TODOs before I think this is ready:
- [ ] fix and support columnGroups
- [ ] fix ember-infinity
- [ ] add missing tests for new functionality
- [ ] reduce duplication (currently a decent amount of copy/paste) to set up the columns
- [ ] figure out a better way to organize the fixed column specific code

screen recording 2016-07-29 at 09 10 pm

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -26,6 +26,13 @@ export default class Column extends Ember.Object.extend({
   hidden: false,
 
   /**
+   * @property fixed
+   * @type {Boolean}
+   * @default false
+   */
+  fixed: false,
+
+  /**
    * @property ascending
    * @type {Boolean}
    * @default true
--- a/addon/classes/Table.js
+++ b/addon/classes/Table.js
@@ -117,7 +117,27 @@ export default class Table extends Ember.Object.extend({
       }
       return arr;
     }, []));
-  }).readOnly()
+  }).readOnly(),
+
+  fixedColumns: computed.filterBy('visibleColumns', 'fixed', true),
+  standardColumns: computed.filterBy('visibleColumns', 'fixed', false),
+
+  fixedColumnWidth: computed('fixedColumns.[]', function() {
+    let fixedColumns = this.get('fixedColumns');
+
+    return fixedColumns.reduce((prev, column) => {
+      if (typeof column.width === 'undefined') {
+        return 0;
+      }
+
+      // column definitions use strings with px, so strip it out
+      let width = parseInt(column.width.replace(/px/, ''));
+
+      return prev + width;
+    }, 0);
+  }),
+
+  //
 }) {
   /**
    * @class Table
--- a/addon/components/columns/base.js
+++ b/addon/components/columns/base.js
@@ -16,7 +16,7 @@ const Column = Component.extend({
   tagName: 'th',
   classNames: ['lt-column'],
   attributeBindings: ['width', 'colspan', 'rowspan'],
-  classNameBindings: ['align', 'isGroupColumn:lt-group-column', 'isHideable', 'isSortable', 'isSorted', 'isResizable', 'column.classNames'],
+  classNameBindings: ['align', 'isGroupColumn:lt-group-column', 'isHideable', 'isSortable', 'isSorted', 'isResizable', 'isFixed', 'column.classNames'],
 
   width: computed.readOnly('column.width'),
   isGroupColumn: computed.readOnly('column.isGroupColumn'),
@@ -24,6 +24,7 @@ const Column = Component.extend({
   isSorted: computed.readOnly('column.sorted'),
   isHideable: computed.readOnly('column.hideable'),
   isResizable: computed.readOnly('column.resizable'),
+  isFixed: computed.readOnly('column.fixed'),
 
   align: computed('column.align', function () {
     return `align-${this.get('column.align')}`;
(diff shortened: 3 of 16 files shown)


offirgolan:
@cball  This is.. this is awesome! Im super excited for this. I think we are gonna have to make the font size on the navbar smaller to fit all these features in 😏 

From a quick glance, I noticed that you're branch is a bit behind master. I would suggest to rebase since it has a lot of changes that resolve some critical issues and that will prolly conflict with some of your code. 

In terms of the trackpad scroll emulation, I think @taras would be your man.

taras:
@cball this is awesome! 

I ended up adding Trackpad Scroll Emulation because we have fixed headers and footers which are setup as separate tables. With regular scrollbars, the columns in the header didn't line up with the body because they had different widths, caused by scrollbar being added to the body.

I'd love a simpler solutions. It'd mean that we'd be able to remove a dependency and reduce complexity. If you have any suggests in regards to how to manage the content width issue caused by body scrollbar, I'd love to hear them.

cball:
Thanks! I'll rebase and get this finished / cleaned up.

In the first pass of this feature, I want to keep the existing virtual scroll behavior in tables without fixed columns. We should be able to normalize the scroll behaviors in all tables with a followup PR. tldr there I think will be measuring scrollbars and explicitly setting widths.

We should also be able to simplify the codebase a bit in the future by making all tables without fixed columns operate the same as the standard columns from the fixed version.

taras:
>  tldr there I think will be measuring scrollbars and explicitly setting widths.

I'm concerned about this. So far, I've been very intentional about not getting into a situation where we're managing dimensions of element through state and rely on the browser to size elements appropriately. 

Mixing state managed sizes and DOM sizes quickly leads to having to manage all sizes through state. This is part of the reason why we're using [`ember-element-resize-detector`](http://github.com/EmberSherpa/ember-element-resize-detector). Instead of forcing nested element to have a height, we're detecting the size of the parent and resizing the child as necessary.

I would prefer for us to continue to evolve this pattern rather than getting into setting widths and heights via state. I'm going to work on a prototype that has nested `ember-scrollable` components. If I can get that to work, then I'll try to build a prototype that shows how we might be able to implement this feature. 

My other concern is that collapsible rows will not work with fixed columns(without doing some complex height management). This is the reason for my comment in #114. I was hoping we'd be able to use something like `ember-collection` where we could create a grid that was fixed columns aware. `ember-collection` never materialized unfortunately.

taras:
It doesn't look like nested `ember-scrollable` approach is going to work. I might have to make `ember-scrollable` support vertical and horizontal scrollbars.

I'll keep you posted on my progress. 

In the meantime, I'm curious what your thoughts are regarding collapsible rows and fixed columns.

nmcclay:
This is a very exciting feature, this behavior could be very useful in regard to expanding the responsive design aspects of this table component.

rwwagner90:
Where does this feature currently stand? I'd be glad to help push it along

cball:
@rwwagner90 I was waiting on feedback on how to proceed. I unfortunately don't have any free cycles at the moment, but if this is a direction that @offirgolan and @taras would like to go with this library (I'm unsure if it is) then help moving things along would be greatly appreciated.

Prabhakar-Poudel:
Is there any progress with the PR? I have been waiting for this feature to be merged.

rwwagner90:
@offirgolan @taras can you guys weigh in here please? I would be glad to help with implementing this, but do not want to put effort in, if it is not something that is desired for this addon.

taras:
This PR has been in a bit of standstill mostly because we're not sure the best way to proceed. We would like to this feature to happen, but we weren't 100% what approach to take. @cball proposed a solution, but we inadvertently blocked him by not providing a path for moving forward.

 @alexander-alvarez has a [PR in the works for `ember-scrollable`](https://github.com/alphasights/ember-scrollable/pull/29) that'll make it possible to have horizontal and vertical scrollbars on a block. Once this lands, it might unblock this PR and provide a way to implement this feature without making the code a lot more complicated.

Would you be interested in picking this up when that PR lands? 

Primarily, we need to figure out how the solution proposed in this PR is affected by that change. `ember-collection` is also becoming much more robust and approaching 1.0. Now might be a good time to see if we can use it to implement this feature.

What do you think?

Prabhakar-Poudel:
has there been there any progress with this?

gspain:
It looks like the ember-scrollable PR has been merged. After a year, is there any news on this feature?

Techn1x:
@cball @taras @offirgolan I'm guessing this PR will probably need a rebase and some reworking, but now that `ember-scrollable` has what is needed, are there any blockers left to getting this PR in?

I think it would be a very popular feature - tables being responsive and usable at small viewports is a huge deal!

PR #143, opened 2016-08-11
[Experimental] Add unfortunate hacks to get ember-sortable working :(

jgwhite:
Ember-sortable is sadly not smart enough to deal with the idea of sorting elements that are not immediately adjacent. The change I had to make to get things working is to mixin sortable-item to columns/base. Not ideal, and a good example of why mixins are usually not the answer :/

Further, because columns/base has its own notion of `width`—which could be a string—I’ve had to clumsily subclass sortable-group to anticipate and parse such values.

All in all, ember-sortable has not proved a particularly elegants solution for this case. Honestly, I’m not surprised… it was written to deal with a very simple case and inevitably users of the addon have come up with much more ambitious solutions.

There is a rewrite underway on the jj-abrams branch of ember-sortable and I’ll definitely take this use case into consideration when designing the new APIs.

For now, I think you may be better served by another solution.

### DIFF
--- a/addon/components/columns/base.js
+++ b/addon/components/columns/base.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import layout from 'ember-light-table/templates/components/columns/base';
+import SortableItem from 'ember-sortable/mixins/sortable-item';
 
 const {
   Component,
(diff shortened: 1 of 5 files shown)


offirgolan:
@jgwhite thanks for taking the time to put this together. I was hoping the solution wouldn't be as messy / hacky. I know @taras has some idea about an alternate route we can take so I will wait for him to proceed. In the mean time, Ill keep my eye out for the next major release of ember-sortable 😺

PR #153, opened 2016-08-28
Add better component example

### LINKED ISSUE #152: Include better examples
Hi there. I've started using this library for the tables in my project and it feels awesome and easy to use. In order to help to improve the library, I have just realised that the proposed example can be improved to a more realistic use case: the code example doesn't updates the light table when the model it's updated in the route or the controller because it's set in the `init` method. Can we add an example updating the Table after the attributes of component are updated in the  `didUpdateAttrs` hook?

Gorzas:
Close #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ const computed = Ember.computed;
 
 export default Ember.Component.extend({
   model: null,
-  table: null,
+  _table: null,
 
   columns: computed(function() {
     return [{
@@ -90,10 +90,19 @@ export default Ember.Component.extend({
     }];
   }),
 
+  table: computed('model.[]', function() {
+    /*
+      Set the table rows when our model changes
+     */
+    this.get('_table').setRows(this.get('model'));
+
+    return this.get('_table');
+  }),
+
   init() {
     this._super(...arguments);
-    this.set('table', new Table(this.get('columns'), this.get('model')));
-  }
+    this.set('_table', new Table(this.get('columns')));
+  },
 });
 ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+## v1.2.0
+
+### Pull Requests
+
+- [#163](https://github.com/offirgolan/ember-light-table/pull/163)  [BUGFIX] Autoprefix addon.css (until PostCSS is up and running) + install ember-cli-autoprefixer to prefix demo page CSS  *by [Offir Golan](https://github.com/offirgolan)*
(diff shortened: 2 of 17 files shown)


offirgolan:
@Gorzas I dont believe that replacing `init` with `didUpdateAttrs` is an ideal implementation (especially in a simple example). The component life cycle hooks `didUpdateAttrs` and `didReceiveAttrs` can fire pretty frequently and can ultimately cause unwanted circumstances for new users that are not aware of its implications.

Gorzas:
Thanks for checking my PR @offirgolan. I didn't replace `init` with `didUpdateAttrs`, I've just added this option. I know it's a simple example but setting `table` only in the `init` method could come to errors in the implementation that are difficult to debug for new users. Can we add it in the example but with a warning?

offirgolan:
@Gorzas `didUpdateAttrs` is not a place you want to keep creating new instances of the table. Instead you can easily use a CP for that and even keep the same table instance

How about something like this instead:

```js
export default Ember.Component.extend({
  model: null,

  columns: computed(function() {
    return [{
      label: 'Avatar',
      valuePath: 'avatar',
      width: '60px',
      sortable: false,
      cellComponent: 'user-avatar'
    }, {
      label: 'First Name',
      valuePath: 'firstName',
      width: '150px'
    }, {
      label: 'Last Name',
      valuePath: 'lastName',
      width: '150px'
    }, {
      label: 'Address',
      valuePath: 'address'
    }, {
      label: 'State',
      valuePath: 'state'
    }, {
      label: 'Country',
      valuePath: 'country'
    }];
  }),

  table: computed('model.[]', function() {
    /*
      Set the table rows when our model changes
     */
    this.get('_table').setRows(this.get('model'));

    return this.get('_table');
  }),

  init() {
    this._super(...arguments);

    /*
      Create our table instance
     */
    this.set('_table', new Table(this.get('columns')));
  }
});
```

Not sure I like using `_table` in the readme though so maybe a better name is appropriate. Maybe this should be a separate example altogether which can show an improved way of doing things on top of the basic example?

In the future, we will implement a way of automatically doing this so there wont be a need for any of this.

README.md
```diff
@@ -90,10 +90,19 @@ export default Ember.Component.extend({
     }];
   }),
 
+  table: computed('model.[]', function() {
```
taras: Computed properties are not supposed to have side effects. If you're going to do this, then you might as well use an observer.

Gorzas:
@offirgolan I have changed the example again, you can take a look. It seems that we may include two examples instead of only one. What do you think?

Note: I have change my commit with a rebase, I don't know if this way is better or if it's better to include an  extra commit instead.

offirgolan:
@Gorzas this will not longer be necessary once #167 lands 😺

Issue #147, opened 2016-08-23
Option to allow row selection without `ctrl`/`cmd`


bargar:
Though I understand the logic behind the `ctrl`-click to toggle individual rows, this may not be intuitive to users in some scenarios. Gmail is one of many examples where no meta keys are required to select/deselect individual rows.

I've added a configuration option to permit this behavior on a fork at [/bargar/ember-light-table](https://github.com/offirgolan/ember-light-table/compare/master...bargar:758a55c).

The existing behavior of course remains the default, but this tweak enables overriding it by setting `lt-body` property `multiSelectRequiresKeyboard` to `false`. Note: the `shift` click to select several consecutive rows behavior is unchanged in any event.

If there is interest and amenability, I would be happy to add testing and make a proper PR here in **/offirgolan/ember-light-table**.

offirgolan:
@bargar a PR with this is very much welcomed 😸

bargar:
@offirgolan #160 submitted, now including testing. Sorry for the delay!

offirgolan:
Thanks again for you contribution @bargar! I should release this sometime this week 😸

bargar:
My pleasure! Looking forward to it 🐸

offirgolan:
@bargar this has been released in `v1.2.0`!

PR #160, opened 2016-09-03
[FEATURE] `multiSelectRequiresKeyboard` option for toggling row selec…

bargar:
…ction without ctrl/cmd

shift-click to select multiple intervening rows works as always

As discussed in #147.

### DIFF
--- a/tests/integration/components/lt-body-test.js
+++ b/tests/integration/components/lt-body-test.js
@@ -24,7 +24,7 @@ test('it renders', function(assert) {
   assert.equal(this.$().text().trim(), '');
 });
 
-test('row selection', function(assert) {
+test('row selection - enable or disable', function(assert) {
   this.set('table', new Table(Columns, createUsers(1)));
   this.set('canSelect', false);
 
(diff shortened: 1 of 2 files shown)


tests/integration/components/lt-body-test.js
```diff
@@ -24,7 +24,7 @@ test('it renders', function(assert) {
   assert.equal(this.$().text().trim(), '');
 });
 
-test('row selection', function(assert) {
+test('row selection - enable or disable', function(assert) {
```
bargar: rather than have three tests called 'row selection' I took the liberty of adding additional detail to these test names to differentiate them

tests/integration/components/lt-body-test.js
```diff
@@ -57,16 +57,40 @@ test('row selection', function(assert) {
...
   middleRow.trigger(createClickEvent({ctrlKey: true}));
-  assert.equal(this.$('tr.is-selected').length, 4);
+  assert.equal(this.$('tr.is-selected').length, 4, 'ctrl-clicking a selected row deselects it');
 
   firstRow.click();
-  assert.equal(this.$('tr.is-selected').length, 0);
+  assert.equal(this.$('tr.is-selected').length, 0, 'clicking a selected row deselects all rows');
+});
+
+test('row selection - click to modify selection', function(assert) {
```
bargar: this new test is very much based on the standard mode test before it, exercising and documenting the differences in behavior for (nearly) identical user input

offirgolan:
@bargar thank you for taking the time in putting this together. Overall, I think the PR looks great! I really appreciate the extra test cases 😸 

Before I merge this, I'd like to pull your branch just so I can get a feel for the feature. I'll prolly do that after this long weekend and we can go from there!

PR #169, opened 2016-09-13
What about initializing table with a computed?

taras:
It allows the table to be overwritten from outside while maintaining same behaviour as setting in init.


offirgolan:
This was added with the `enableSync` PR

PR #186, opened 2016-09-25
[DOCS] fix confusing property names for lt-body

acburdine:
no issue
- the three different types of row components were all named the same in the documentation


offirgolan:
@acburdine good catch! Thank you 😄

PR #193, opened 2016-09-30
Send row selection state changes to the onRowClick handler

billdami:
This is my stab at one of the changes I proposed in #151.  The onRowClick action will now send an array of rows whose selection state has been changed as its third parameter. This should allow for easier handling of row selection changes that originate from row clicks, instead of having to setup an observer on `table.selectedRows`, especially if you need to persist the state changes externally, such as to a backend.

### DIFF
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -176,7 +176,7 @@ export default Component.extend({
 
   /**
    * Allows to customize the component used to render rows
-   * 
+   *
    * ```hbs
    * {{#light-table table as |t|}}
    *    {{t.body rowComponent=(component 'my-row')}}
@@ -190,7 +190,7 @@ export default Component.extend({
 
   /**
    * Allows to customize the component used to render spanned rows
-   * 
+   *
    * ```hbs
    * {{#light-table table as |t|}}
    *    {{t.body spannedRowComponent=(component 'my-spanned-row')}}
@@ -204,7 +204,7 @@ export default Component.extend({
 
   /**
    * Allows to customize the component used to render infinite loader
-   * 
+   *
    * ```hbs
    * {{#light-table table as |t|}}
    *    {{t.body infinityComponent=(component 'my-infinity')}}
@@ -257,6 +257,7 @@ export default Component.extend({
      * @event onRowClick
      * @param  {Row}   row The row that was clicked
      * @param  {Event}   event   The click event
+     * @param  {Array}   affectedSelections   The rows whose selection state has been changed
      */
     onRowClick(row, e) {
       let rows = this.get('table.rows');
@@ -266,17 +267,31 @@ export default Component.extend({
       let isSelected = row.get('selected');
       let currIndex = rows.indexOf(row);
       let prevIndex = this._prevSelectedIndex === -1 ? currIndex : this._prevSelectedIndex;
+      let affectedSelections = Ember.A();
 
       this._currSelectedIndex = currIndex;
       this._prevSelectedIndex = prevIndex;
 
       if (canSelect) {
         if (e.shiftKey && multiSelect) {
-          rows.slice(Math.min(currIndex, prevIndex), Math.max(currIndex, prevIndex) + 1).forEach(r => r.set('selected', !isSelected));
+          rows.slice(Math.min(currIndex, prevIndex), Math.max(currIndex, prevIndex) + 1).forEach(r => {
+            if (!r.get('selected')) {
+                affectedSelections.pushObject(r);
+            }
+
+            r.set('selected', !isSelected);
+          });
           this._prevSelectedIndex = currIndex;
         } else if ((!multiSelectRequiresKeyboard || (e.ctrlKey || e.metaKey)) && multiSelect) {
           row.toggleProperty('selected');
+          affectedSelections.pushObject(row);
         } else {
+          if (!isSelected) {
+            affectedSelections.pushObject(row);
+          }
+
+          affectedSelections.pushObjects(this.get('table.selectedRows'));
+
           this.get('table.selectedRows').setEach('selected', false);
           row.set('selected', !isSelected);
 
@@ -291,7 +306,7 @@ export default Component.extend({
         }
       }
 
-      callAction(this, 'onRowClick', ...arguments);
+      callAction(this, 'onRowClick', row, e, affectedSelections);
     },
 
     /**
--- a/tests/integration/components/lt-body-test.js
+++ b/tests/integration/components/lt-body-test.js
@@ -149,12 +149,18 @@ test('row expansion - multiple', function(assert) {
 });
 
 test('row actions', function(assert) {
-  assert.expect(2);
+  assert.expect(3);
 
   this.set('table', new Table(Columns, createUsers(1)));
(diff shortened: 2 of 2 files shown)


addon/components/lt-body.js
```diff
@@ -266,17 +267,31 @@ export default Component.extend({
       let isSelected = row.get('selected');
       let currIndex = rows.indexOf(row);
       let prevIndex = this._prevSelectedIndex === -1 ? currIndex : this._prevSelectedIndex;
+      let affectedSelections = Ember.A();
```
offirgolan: I dont think you need to use an Ember.Array here

addon/components/lt-body.js
```diff
@@ -266,17 +267,31 @@ export default Component.extend({
...
           this._prevSelectedIndex = currIndex;
         } else if ((!multiSelectRequiresKeyboard || (e.ctrlKey || e.metaKey)) && multiSelect) {
           row.toggleProperty('selected');
+          affectedSelections.pushObject(row);
         } else {
+          if (!isSelected) {
+            affectedSelections.pushObject(row);
+          }
+
+          affectedSelections.pushObjects(this.get('table.selectedRows'));
```
offirgolan: Why are you adding all the selectedRows here?

addon/components/lt-body.js
```diff
@@ -291,7 +306,7 @@ export default Component.extend({
         }
       }
 
-      callAction(this, 'onRowClick', ...arguments);
+      callAction(this, 'onRowClick', row, e, affectedSelections);
```
offirgolan: `callAction(this, 'onRowClick', ...arguments, affectedSelections);`

addon/components/lt-body.js
```diff
@@ -266,17 +267,31 @@ export default Component.extend({
...
+
+            r.set('selected', !isSelected);
+          });
           this._prevSelectedIndex = currIndex;
         } else if ((!multiSelectRequiresKeyboard || (e.ctrlKey || e.metaKey)) && multiSelect) {
           row.toggleProperty('selected');
+          affectedSelections.pushObject(row);
         } else {
+          if (!isSelected) {
+            affectedSelections.pushObject(row);
```
offirgolan: Why is this only pushed if `isSelected` is false? 

No matter what the situation is, if `canSelect` is true, the target row will be affected. I really dont think we need to add it to the list. What if we generalized it a bit more?

What if affectedSelections becomes a collection of other rows that were affected by the click event? This can include both `expanded` and `selected`.

offirgolan: [restarting review...]

addon/components/lt-body.js
billdami: I was originally going to use a native array here, but w/o prototype extensions turned on, I'm unable to use ember's convenience methods such as `pushObjects()`.  Also since this array is being sent to component consumer via the action, I wanted to make sure they are always getting a nice Ember array regardless of their environment configuration.

addon/components/lt-body.js
offirgolan: > To avoid adding the target row to the affectedSelections array twice

Like I said, we should be adding ALL rows that have changed EXCLUDING the current row. So you can return something like this:

```js
affectedSelections.uniq().without(currentRow)
```

>  I think it might be better to send them as a separate array parameter, e.g. affectedRowExpansions or something like that

What if we send a `changes` hash?

```js
{
  selected: [ ],
  expanded: [ ]
}
```

addon/components/lt-body.js
offirgolan: 👍

addon/components/lt-body.js
billdami: @offirgolan sounds good to me. I've pushed those changes.

addon/components/lt-body.js
```diff
@@ -176,7 +176,7 @@ export default Component.extend({
```
offirgolan: Add `A: emberArray` to the `const` at the top of the file

addon/components/lt-body.js
```diff
@@ -242,13 +242,17 @@ export default Component.extend({
   toggleExpandedRow(row) {
     let multi = this.get('multiRowExpansion');
     let shouldExpand = !row.expanded;
+    let otherAffectedRows = Ember.A();
```
offirgolan: `let affectedRows = []`

addon/components/lt-body.js
```diff
@@ -242,13 +242,17 @@ export default Component.extend({
...
 
     if (multi) {
       row.toggleProperty('expanded');
     } else {
+      otherAffectedRows.pushObjects(this.get('table.expandedRows'));
```
offirgolan: `affectedRows.push`

addon/components/lt-body.js
```diff
@@ -257,6 +261,7 @@ export default Component.extend({
      * @event onRowClick
      * @param  {Row}   row The row that was clicked
      * @param  {Event}   event   The click event
+     * @param  {Array}   affectedSelections   The rows whose selection state has been changed
```
offirgolan: {Object} affectedRows

addon/components/lt-body.js
```diff
@@ -266,32 +271,46 @@ export default Component.extend({
       let isSelected = row.get('selected');
       let currIndex = rows.indexOf(row);
       let prevIndex = this._prevSelectedIndex === -1 ? currIndex : this._prevSelectedIndex;
+      let otherAffectedRows = {
```
offirgolan: `let affectedRows = { selected: [] }`

addon/components/lt-body.js
```diff
@@ -266,32 +271,46 @@ export default Component.extend({
...
 
       this._currSelectedIndex = currIndex;
       this._prevSelectedIndex = prevIndex;
 
       if (canSelect) {
         if (e.shiftKey && multiSelect) {
-          rows.slice(Math.min(currIndex, prevIndex), Math.max(currIndex, prevIndex) + 1).forEach(r => r.set('selected', !isSelected));
+          rows.slice(Math.min(currIndex, prevIndex), Math.max(currIndex, prevIndex) + 1).forEach(r => {
+            if (!r.get('selected')) {
+                otherAffectedRows.selected.pushObject(r);
```
offirgolan: `affectedRows.selected.push`

addon/components/lt-body.js
```diff
@@ -266,32 +271,46 @@ export default Component.extend({
...
         this._prevSelectedIndex = currIndex;
       } else {
         if (this.get('canExpand') && this.get('expandOnClick')) {
-          this.toggleExpandedRow(row);
+          otherAffectedRows.expanded = this.toggleExpandedRow(row);
         }
       }
 
-      callAction(this, 'onRowClick', ...arguments);
+      otherAffectedRows.selected = otherAffectedRows.selected.uniq().without(row);
```
offirgolan: `affectedRows.selected = emberArray(affectedRows.selected).uniq().without(row)`

same for expanded

addon/components/lt-body.js
offirgolan: Also, since we are doing `uniq` at the end, is this if statement needed?

PR #194, opened 2016-09-30
Proxyable row options

billdami:
An attempt at solving the problem discussed in #151, allowing row options/state properties (notably, the `selected` state) to be proxied to the row's content object. However, this PR feels kinda messy/ugly and probably is not the best implementation, so its more intended as a way to kick off further discussion and ideas.

A major thing to point out is that I ended up requiring a global option, `proxyRowOptions` to be enabled in order to opt-in to the proxying of the row options. I did this because I think without it, this would be a potentially breaking change to the API, since setting the row's `hidden`, `expanded`, and `selected` properties are now updating the row's content object itself, which may already contain properties with the same name that are used for other purposes.

### DIFF
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -1,10 +1,11 @@
 import Ember from 'ember';
+import globalOptions from 'ember-light-table/-private/global-options';
 
  /**
   * @module Table
   * @class Row
   */
-export default class Row extends Ember.ObjectProxy.extend({
+class Row extends Ember.ObjectProxy.extend({
   /**
    * @property hidden
    * @type {Boolean}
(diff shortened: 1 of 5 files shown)


offirgolan:
@billdami thank you for submitting this PR! Although I understand the use-case for this, I dont believe this is the correct solution. In all honesty, with the current implementation, I dont think there is a good solution for this. This could be one of those things that you will have to go around until we can figure out a decent solution. 

@taras I would love your opinion on this one.

billdami:
@offirgolan No problem! Yeah it definitely didn't feel like the best solution as I was implementing it, but as I mentioned in the OP, I more intended it to be a jumping off point for some more serious discussion on how it should be done. :)

offirgolan:
@billdami I'm glad you feel the same way! I totally agree with you on the fact that this needs some much needed discussion. As of right now, there are some bigger features that we are trying to push for so we will have to place this issue on the backlog until we have some more time. Thanks again for taking the time to put this together! 😸

Issue #198, opened 2016-10-03
Inconsistent Doc: lt-head


alexander-alvarez:
why are there three arguments being yielded here `|groups subColumns table|`,
https://github.com/offirgolan/ember-light-table/blob/master/addon/components/lt-head.js#L27

but only two here?
https://github.com/offirgolan/ember-light-table/blob/master/addon/templates/components/lt-head.hbs#L5

offirgolan:
@alexander-alvarez seems like I forgot to remove the table yield from the docs. My apologies. A PR with the fix is welcomed 😄

alexander-alvarez:
np. wondering if there was something else that was intended

PR #200, opened 2016-10-04
Docfix: lt-head and column-type blueprint

alexander-alvarez:
closes https://github.com/offirgolan/ember-light-table/issues/198


offirgolan:
👍 Thank you! 😺

PR #202, opened 2016-10-05
(docs): Namespacing the classes

alexander-alvarez:
Looks like so.
image

That way people can import: http://offirgolan.github.io/ember-light-table/docs/data.json into their yuidoc.json externals, and not get their classes clobbered.

Wondering if I should namespace all the components?


alexander-alvarez:
@offirgolan thoughts?

PR #221, opened 2016-10-12
[FEATURE] Add label to base column to be used in column types

offirgolan:
This puts `label` on the base column so it can be easily overwritten when defining new column types.


HenryVonfire:
It worked like a charm with `ember-intl` :smile: 

```js
export default Column.extend({
  intl: service(),
  label: computed('column.label', function () {
    return this.get('intl').findTranslationByKey(this.get('column.label'));
  })
});
```

PR #222, opened 2016-10-12
[FEATURE] Support horizontal scrolling

offirgolan:
Conditions:
- All widths must be defined on all visible columns
- Widths must be the same unit


offirgolan:
@taras I would love your feedback on this one

taras:
How complete is this implementation? I guess it just uses regular scrollbar?

offirgolan:
Its very naive. It works well but just uses the native scrollbar.

addon/components/lt-infinity.js
```diff
@@ -14,13 +14,17 @@ export default Component.extend(InViewportMixin, {
 
   didInsertElement() {
     this._super(...arguments);
+
+    let scrollBuffer = this.get('scrollBuffer');
```
offirgolan: Added Ember Suave to fix this kind of issues. It says to use `let` so I'm going with it 😝

alexander-alvarez: not sure what it's deal is, but there's so many variables that aren't being reassigned, so should be const
https://github.com/dockyard/styleguides/blob/master/engineering/javascript.md#assignment

gwak:
Hi, any idea when this PR going to be merged into master ?

offirgolan:
I've been waiting on @alexander-alvarez to see if he can find a workable solution using ember-scrollable. Maybe I can release this as a temporary solution until then.

alexander-alvarez:
progress has been made... about to attempt to do the double scrolling - https://github.com/alphasights/ember-scrollable/pull/29

offirgolan:
@alexander-alvarez thats great to hear! Looking at the PR, it seems like your solution will take a bit of work and some time until it gets reviewed / merged. I think Im gonna go ahead and merge this so that way there is no real pressure to get a final solution in.

gwak:
Thanks, it works great and sufficient for our need right now. I look forward to check ember-scrollable solution though.

gwak:
Actually what I eventually did is setup an event handler for scroll on the body scrollable-content div with jQuery in the didRender hook of my component.  I then simply modify the scrollLeft of the .lt-head-wrap to match the scrollLeft of the ".lt-body-wrap .scrollable-content" . It works pretty well, be you need to enable overflow-x:auto on the body content and set an id to the light-table component.

meliborn:
Does it work? I have header and footer fixed both and long header, but I can't scroll and they are overlaping each other.

offirgolan:
@meliborn yes its works. [See here](http://offirgolan.github.io/ember-light-table/#/cookbook/horizontal-scrolling)

offirgolan:
@meliborn Make sure all your columns specify a `width` and all width are the same unit.

gwak:
Hi @offirgolan, the horizontal scroll works fine when the element containing the table is narrower than the table, but as soon as the container is wider than the table the table doesn't fill up the container. I guess it would be nice if the table width matches the container width if the container is wider than the table width.
Would it be possible ?

meliborn:
@offirgolan What should I do to enable horizontal scrolling? Can't find any specific code in provided example.

meliborn:
Had to add .table-container. But there is a problem with pagination - it scrolls as well, even with fixed=true.

offirgolan:
@meliborn can you open a new issue with some screenshots and maybe a reproduction?

Issue #224, opened 2016-10-12
dummy/snippets: store service not injected, RecordArray mutated


buschtoens:
[The snippets](https://github.com/offirgolan/ember-light-table/tree/master/tests/dummy/snippets) in the dummy app assume that the `store` service is [injected](https://guides.emberjs.com/v2.8.0/applications/initializers/#toc_application-initializers) into all `Component`s just as with `Route`s, which it apparently isn't. This can confuse beginners as seen in #223.

Suggested changes:
#### Manually inject the store at the top of every snippet.

``` js
export default Ember.Component.extend({
  store: Ember.inject.service(),

  page: 1,
  // ...
});
```
#### Use `this.get('store')` instead of `this.store`.

``` js
  fetchRecords() {
    this.set('isLoading', true);
    this.get('store').query('user', this.getProperties(['page', 'limit', 'sort', 'dir'])).then(records => {
      this.get('model').pushObjects(records.toArray());
      this.set('isLoading', false);
      this.set('canLoadMore', !isEmpty(records));
    });
  },
```

---

Addendum: The original `RecordArray` is immutable. Currently we use `pushObjects` on the original `RecordArray` returned from the `model` hook. This shouldn't work, and breaks users apps in the wild.

Suggested changes:
#### Use a private `_model` that is a `model.toArray()` clone.

``` js
model: null,
_model: Ember.computed('model', function() {
  return this.get('model').toArray();
})
```

_...or..._
#### Don't use `enableSync: true` in all examples, but a special one

---

I'm down for submitting a PR.

offirgolan:
@buschtoens good catch. PR is totally welcomed!

trumb1mj:
When using this in a component I get `Uncaught TypeError: internalModel.getRecord is not a function(…)`

Anyone seeing this?

trumb1mj:
How I ended up solving this was using the a private attribute like so:

```
  fetchRecords() {
    this.set('isLoading', true);
    this.get("store").query('account', {per_page: this.get("limit"), page: this.get("page")}).then(records => {
      var model = this.get('model');
      records.forEach(function(record){
        model.addObject(record._internalModel);
      });
      this.set('isLoading', false);
      this.set('canLoadMore', !isEmpty(records));
    });
  },
```

buschtoens:
Interesting. We already had this error in #223...

Which versions of ember-light-table, Ember.js and ember-data are you using?

trumb1mj:
[...]

PR #228, opened 2016-10-18
[FEATURE] ScrollTo and ScrollToRow

### LINKED ISSUE #148: Scrolling a row into view
```js
 /**
  * @module Table
  * @class Table
  */
export default class Table extends Ember.Object.extend({

  // ...

  /**
   * Scrolls the row with the specified index to the top of the viewport.
   * @method scrollToRow
   * @param  {Number} index
   * @return {Row} target row
   */
  scrollToRow(index = 0) {
    // insert magic here
  }
});
```

I'd really like this feature and would be down to submit a PR, if you like it as well. :)

buschtoens:
Closes #148.

### DIFF
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-const { computed, isBlank } = Ember;
+const { computed, guidFor } = Ember;
 
  /**
   * @module Table
@@ -37,27 +37,15 @@ export default class Row extends Ember.ObjectProxy.extend({
   classNames: null,
 
   /**
-   * @property rowComponent
-   * @type {Ember.Component}
-   * @private
-   */
-  rowComponent: null,
-
-  /**
-   * @property scrollOffset
-   * @type {Number}
-   * @private
+   * Element ID for the corresponding `{{lt-row}}`.
+   *
+   * @property elementId
+   * @type {String}
    * @readOnly
    */
-  scrollOffset: computed(function() {
-    const element = this.get('rowComponent.element');
-
-    if (isBlank(element)) {
-      return null;
-    }
-
-    return element.offsetTop;
-  }).readOnly().volatile()
+  elementId: computed(function() {
+    return `lt-row-${guidFor(this)}`;
+  }).readOnly()
 }) {
   /**
    * @class Row
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -2,10 +2,10 @@ import Ember from 'ember';
 
 const { computed, guidFor } = Ember;
 
- /**
-  * @module Table
-  * @class Row
-  */
+/**
+ * @module Table
+ * @class Row
+ */
 export default class Row extends Ember.ObjectProxy.extend({
   /**
    * @property hidden
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -1,9 +1,11 @@
 import Ember from 'ember';
 
-/**
- * @module Table
- * @class Row
- */
+const { computed, isBlank } = Ember;
+
+ /**
+  * @module Table
+  * @class Row
+  */
 export default class Row extends Ember.ObjectProxy.extend({
   /**
    * @property hidden
@@ -32,7 +34,30 @@ export default class Row extends Ember.ObjectProxy.extend({
    * @property classNames
    * @type {String | Array}
    */
-  classNames: null
+  classNames: null,
+
+  /**
+   * @property rowComponent
+   * @type {Ember.Component}
+   * @private
+   */
+  rowComponent: null,
+
+  /**
+   * @property scrollOffset
+   * @type {Number}
+   * @private
+   * @readOnly
+   */
+  scrollOffset: computed(function() {
+    const element = this.get('rowComponent.element');
+
+    if (isBlank(element)) {
+      return null;
+    }
+
+    return element.offsetTop;
+  }).readOnly().volatile()
 }) {
   /**
    * @class Row
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -39,12 +39,14 @@ export default class Row extends Ember.ObjectProxy.extend({
   /**
    * Element ID for the corresponding `{{lt-row}}`.
    *
-   * @property elementId
+   * Note: named `rowId` in order to not shadow the `content.id` property.
+   *
+   * @property rowId
    * @type {String}
    * @readOnly
    */
-  elementId: computed(function() {
-    return `lt-row-${guidFor(this)}`;
+  rowId: computed(function() {
+    return guidFor(this);
   }).readOnly()
 }) {
   /**
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -8,7 +8,8 @@ const {
   computed,
   run,
   observer,
-  get
+  get,
+  isPresent
 } = Ember;
 
 /**
@@ -298,25 +299,49 @@ export default Component.extend({
     let newScrollToRow = get(newAttrs, 'scrollToRow.value');
     let oldScrollToRow = get(oldAttrs, 'scrollToRow.value');
 
-    let targetScrollOffset = 0;
-    let hasReachedTargetScrollOffset = true;
-
     if (oldScrollTo !== newScrollTo) {
-      targetScrollOffset = newScrollTo;
-      if (newScrollTo > 0) {
-        hasReachedTargetScrollOffset = false;
-      }
+      this._scrollTo(newScrollTo);
     } else if (oldScrollToRow !== newScrollToRow) {
-      if (newScrollToRow instanceof Row) {
-        let rowElement = document.getElementById(newScrollToRow.get('rowId'));
-        if (rowElement instanceof Element) {
-          targetScrollOffset = rowElement.offsetTop;
-        }
+      this._scrollToRow(newScrollToRow);
+    }
+  },
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    let scrollTo = this.get('scrollTo');
+    let scrollToRow = this.get('scrollToRow');
+
+    if (isPresent(scrollTo)) {
+      this._scrollTo(scrollTo);
+    } else if (isPresent(scrollToRow)) {
+      this._scrollToRow(scrollToRow);
+    }
+  },
+
+  _scrollTo(offset) {
+    let targetScrollOffset = Number.parseInt(offset, 10);
+
+    if (Number.isNaN(targetScrollOffset)) {
+      targetScrollOffset = null;
+    }
+
+    this.set('targetScrollOffset', targetScrollOffset);
+    this.set('hasReachedTargetScrollOffset', targetScrollOffset <= 0);
+  },
+
+  _scrollToRow(row) {
+    let targetScrollOffset = null;
+
+    if (row instanceof Row) {
+      let rowElement = document.getElementById(row.get('rowId'));
+      if (rowElement instanceof Element) {
+        targetScrollOffset = rowElement.offsetTop;
       }
     }
 
     this.set('targetScrollOffset', targetScrollOffset);
-    this.set('hasReachedTargetScrollOffset', hasReachedTargetScrollOffset);
+    this.set('hasReachedTargetScrollOffset', true);
   },
 
   destroy() {
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -6,7 +6,8 @@ import Row from 'ember-light-table/classes/Row';
 const {
   Component,
   computed,
-  run
+  run,
+  observer
 } = Ember;
 
 /**
@@ -208,6 +209,22 @@ export default Component.extend({
   targetScrollOffset: null,
 
   /**
+   * @property currentScrollOffset
+   * @type {Number}
+   * @default null
+   * @private
+   */
+  currentScrollOffset: true,
+
+  /**
+   * @property hasReachedTargetScrollOffset
+   * @type {Boolean}
+   * @default true
+   * @private
+   */
+  hasReachedTargetScrollOffset: true,
+
+  /**
    * Allows to customize the component used to render rows
    *
    * ```hbs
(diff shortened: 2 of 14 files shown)


buschtoens:
I pushed [a build to my `gh-pages`](https://buschtoens.github.io/ember-light-table/#/scrolling) for your convenience.

Are you happy with the way the demo and actual code are implemented or would you like me to make changes?

buschtoens:
This PR deliberately ignores the issues raised in #224. I'll tackle them in one go when this is merged.

buschtoens:
Something worth noting:

In order for the `showRow` feature to work, I needed a way to retrieve the [`element.offsetTop`](https://github.com/offirgolan/ember-light-table/pull/228/commits/e9275c536f5f78c881847ce3819ab7f31faa40e4#diff-e6209e612f962248f43f79115282ae06R59) property of the corresponding `{{lt-row}}` component for a given `Row` instance.

Currently the `{{lt-row}}` component has a reference to the `Row` instance, but not the other way around. I therefore added the private [`rowComponent` property](https://github.com/offirgolan/ember-light-table/pull/228/commits/e9275c536f5f78c881847ce3819ab7f31faa40e4#diff-e6209e612f962248f43f79115282ae06R44) to the `Row` class which is populated in the [`didInsertElement` hook](https://github.com/offirgolan/ember-light-table/pull/228/commits/e9275c536f5f78c881847ce3819ab7f31faa40e4#diff-ae1775e32e134fff961683c296123414R26) of the `{{lt-row}}` component.

I feel like this is the cleanest way. But I'm up for discussion. An alternative would be getting the `this.get('table.rows').indexOf(targetRow)` and querying the DOM for the n-th row.

buschtoens:
Oh, and the "keep fetching rows until the `targetScrollOffset` is reached" feature isn't implemented yet.

offirgolan:
@buschtoens thank you for taking the time in putting this together. I do have a few concerns. 

1. Offset logic should not be placed in the `Row` class. 
2. What happens when we are not using ember-scrollable (i.e. simple use cases with no fixed header/footer?
3. Why do we need to keep fetching rows until the target row is reached? If we require the Row instance to exist, shouldn't that instance already be in the table / rendered? 

In regards to my first concern, I believe we can use the row instance's guid as the id of the lt-row component instance'd id. With that, we can get the offset of the row from the lt-body. 

```js
id: computed(function() {
  return guidFor(this);
}).readOnly()
```

```hbs
{{lt-row id=row.id row=row ...}}
```

buschtoens:
Thanks for replying!

> 1) Offset logic should not be placed in the `Row` class.

I agree with that. Using the GUID is a neat idea.

> 3) Why do we need to keep fetching rows until the target row is reached? If we require the Row instance to exist, shouldn't that instance already be in the table / rendered?

This is true when using the `showRow` interface. But there's also a `scrollTo` interface which just takes a numeric scroll offset in px. Primarily I'm interested in the latter one to easily restore state after changing the route.

> 2) What happens when we are not using ember-scrollable (i.e. simple use cases with no fixed header/footer?

So far, nothing. As far as I'm concerned I only need these features when I have a table that is intended to overflow and scroll, in which case I'm already opting for the fixed headers/footers.

I can't think of a legit reason to not use fixed headers/footers, other than possible performance gains, because the UX of disappearing headers/footers is pretty moot.

Options:

a. Explicitly document that `showRow`, `scrollTo` and `onScroll` only work when using fixed headers/footers.
b. Basically reimplement https://github.com/alphasights/ember-scrollable/pull/25, however this only works when the user's CSS targets the `{{lt-body}}` as the scroll container.

offirgolan:
> This is true when using the showRow interface. But there's also a scrollTo interface which just takes a numeric scroll offset in px. Primarily I'm interested in the latter one to easily restore state after changing the route.

Ah right. I forgot about that use case.

> Explicitly document that showRow, scrollTo and onScroll only work when using fixed headers/footers.

I think this is okay 😄

offirgolan:
@buschtoens dont worry too much on the demo. When this is fully functional and merged, I'll go back and make some changes. I also have a client that will require this feature so if you need any help, just let me know or shoot me a message on slack.

buschtoens:
@offirgolan Ready for review. :)

addon/classes/Row.js
```diff
@@ -33,6 +35,17 @@ export default class Row extends Ember.ObjectProxy.extend({
...
   classNames: null,
+
+  /**
+   * Element ID for the corresponding `{{lt-row}}`.
+   *
+   * @property elementId
+   * @type {String}
+   * @readOnly
+   */
+  elementId: computed(function() {
```
offirgolan: Lets keep this as `id` and just return the guid. This should be unique enough. I just dont think any property here should directly correlate to a component.

addon/components/lt-body.js
```diff
@@ -175,8 +177,56 @@ export default Component.extend({
...
+   * Set this property to a `Row` to scroll that `Row` into view.
+   *
+   * This only works when `useVirtualScrollbar` is `true`, i.e. when you are
+   * using fixed headers / footers.
+   *
+   * @property showRow
+   * @type {Row}
+   * @default null
+   */
+  showRow: null,
```
offirgolan: I think `scrollToRow` is a more fitting name here and is consistent with the other `scrollTo` property.

addon/components/lt-body.js
```diff
@@ -239,6 +292,57 @@ export default Component.extend({
     this.set('useVirtualScrollbar', fixedHeader || fixedFooter);
   },
 
+  didReceiveAttrs() {
```
offirgolan: Would it be better to handle this with `didUpdateAttrs` since it gives you the before and after values? That could remove the need of having `_scrollTo` and `_scrollToRow`. You can handle the initial scroll with `didInsertElement`

addon/components/lt-body.js
```diff
@@ -239,6 +292,57 @@ export default Component.extend({
...
+    const showRow = this.get('showRow');
+    const _showRow = this.get('_showRow');
+
+    this.set('_scrollTo', scrollTo);
+    this.set('_showRow', showRow);
+
+    if (scrollTo !== _scrollTo) {
+      this.set('targetScrollOffset', scrollTo);
+      this.set('hasReachedTargetScrollOffset', false);
+      run.scheduleOnce('afterRender', this, this.checkTargetScrollOffset);
```
offirgolan: I noticed this is done in more than one place. Lets move this to another method. 

Also, please make sure to capture all timer instances and cancel them destroy in `destroy`.

addon/components/lt-body.js
```diff
@@ -175,8 +177,56 @@ export default Component.extend({
...
+   */
+  showRow: null,
+
+  /**
+   * @property targetScrollOffset
+   * @type {Number}
+   * @default null
+   * @private
+   */
+  targetScrollOffset: null,
```
offirgolan: Set the default here to `0`

addon/components/lt-body.js
```diff
@@ -175,8 +177,56 @@ export default Component.extend({
...
+
+  /**
+   * @property currentScrollOffset
+   * @type {Number}
+   * @default null
+   * @private
+   */
+  currentScrollOffset: true,
```
offirgolan: Set the default here to `0`

addon/components/lt-body.js
```diff
@@ -175,8 +177,56 @@ export default Component.extend({
...
+
+  /**
+   * @property hasReachedTargetScrollOffset
+   * @type {Boolean}
+   * @default true
+   * @private
+   */
+  hasReachedTargetScrollOffset: true,
```
offirgolan: Can this be a CP? Something like:

`hasReachedTargetScrollOffset: computed.gte('currentScrollOffset', 'targetScrollOffset');`

addon/templates/components/lt-body.hbs
```diff
@@ -18,6 +20,7 @@
         {{else}}
           {{#each rows as |row|}}
             {{lt.row row columns
+              elementId=row.elementId
```
offirgolan: id=row.id

addon/classes/Row.js
offirgolan: Note: The `Column` class will also have the same implementation once I finish the drag n drop column headers feature.

addon/components/lt-body.js
offirgolan: Also, none of this should be happening if `scrollTo` and `scrollToRow` are not set.

addon/classes/Row.js
buschtoens: I named this field `elementId`, so it does not shadow the `id` field of the `content` object which is most likely an Ember Data `Record`.

Still want me to rename it?

addon/components/lt-body.js
buschtoens: I agree.

addon/components/lt-body.js
buschtoens: This would trigger a scroll to `0 px` down the line. However that shouldn't matter as we start at `0 px` anyway. So for clarity's sake, you're probably right.

addon/components/lt-body.js
buschtoens: Yeah, that was a copy pasta typo.

addon/components/lt-body.js
buschtoens: I didn't know that `didUpdateAttrs()` actually can be used to diff the attrs. Nice one, thanks!

addon/templates/components/lt-body.hbs
buschtoens: Actually *has* to be named [`elementId`](http://emberjs.com/api/classes/Ember.Component.html#property_elementId).

addon/components/lt-body.js
buschtoens: Side note: `checkTargetScrollOffset()` is only applicable for `scrollTo`. Other than that I agree.

addon/templates/components/lt-body.hbs
offirgolan: Ah gotcha. Thanks for clarifying. I believe the `id` field is an alias to the `elementId`

addon/classes/Row.js
offirgolan: Ah thats a good call. Maybe change this to `rowId` and add a small comment as to why we didn't go by `id` (just so I randomly change this later 😛 )

addon/components/lt-body.js
```diff
@@ -175,6 +178,54 @@ export default Component.extend({
...
+   */
+  scrollToRow: null,
+
+  /**
+   * @property targetScrollOffset
+   * @type {Number}
+   * @default 0
+   * @private
+   */
+  targetScrollOffset: 0,
```
buschtoens: Due to this property triggering an initial scroll, this PR is blocked until https://github.com/alphasights/ember-scrollable/pull/26 gets published.

offirgolan: cc @taras

addon/components/lt-body.js
```diff
@@ -239,6 +290,55 @@ export default Component.extend({
...
+    this.set('targetScrollOffset', targetScrollOffset);
+    this.set('hasReachedTargetScrollOffset', hasReachedTargetScrollOffset);
+  },
+
+  rowObserver: observer('rows.[]', function() {
+    run.scheduleOnce('afterRender', this, this.checkTargetScrollOffset);
+  }),
+
+  checkTargetScrollOffset() {
+    if (!this.get('hasReachedTargetScrollOffset') && !this.get('isDestroyed')) {
```
buschtoens: I'm checking for `isDestroyed` instead of cancelling the queued task, which effectively achieves the same thing. Are you okay with this?

offirgolan: I dont think this is enough. On line 334 you are doing a `set` on what could be a destroyed object. I think capturing all viable timers and cancelling them on destroy is a much safer bet.

addon/components/lt-body.js
```diff
@@ -239,6 +290,55 @@ export default Component.extend({
     this.set('useVirtualScrollbar', fixedHeader || fixedFooter);
   },
 
+  didUpdateAttrs({ newAttrs, oldAttrs }) {
+    this._super(...arguments);
+
+    let newScrollTo = get(newAttrs, 'scrollTo.value');
+    let oldScrollTo = get(oldAttrs, 'scrollTo.value');
+    let newScrollToRow = get(newAttrs, 'scrollToRow.value');
+    let oldScrollToRow = get(oldAttrs, 'scrollToRow.value');
```
offirgolan: Huh... you're totally right. That functionality is no longer documented. Lets revert back to your the previous implementation.

addon/components/lt-body.js
```diff
@@ -239,6 +290,55 @@ export default Component.extend({
...
+        }
+      }
+    }
+
+    this.set('targetScrollOffset', targetScrollOffset);
+    this.set('hasReachedTargetScrollOffset', hasReachedTargetScrollOffset);
+  },
+
+  rowObserver: observer('rows.[]', function() {
+    run.scheduleOnce('afterRender', this, this.checkTargetScrollOffset);
```
offirgolan: @buschtoens I suggest adding a private method:

```js
_scheduleCheckTargetScrollOffset() {
  run.cancel(this._checkTargetOffsetTimer); // Cancel existing checks since only one needs to happen (I'm not sure if this is actually needed since we are using scheduleOnce.... but just to be safe?
  this._checkTargetOffsetTimer = run.scheduleOnce('afterRender', this, this.checkTargetScrollOffset);
}

destroy() {
  this._super(...arguments);
  run.cancel(this._checkTargetOffsetTimer);
}
```

buschtoens: This wouldn't account for the timer in line 334. But tbh, I doubt that the timer in 334 could ever cause problems.

---

`run.cancel(...)` inside `_scheduleCheckTargetScrollOffset()` is unnecessary.

Relevant code in Backburner:
https://github.com/ebryn/backburner.js/blob/master/lib/backburner/deferred-action-queues.js#L38-L42
https://github.com/ebryn/backburner.js/blob/master/lib/backburner/queue.js#L33-L37

buschtoens: So should I add a second reference for that timer as well? :smile:

addon/components/lt-body.js
```diff
@@ -239,6 +290,61 @@ export default Component.extend({
...
+
+    let targetScrollOffset = 0;
+    let hasReachedTargetScrollOffset = true;
+
+    if (oldScrollTo !== newScrollTo) {
+      targetScrollOffset = newScrollTo;
+      if (newScrollTo > 0) {
+        hasReachedTargetScrollOffset = false;
+      }
+    } else if (oldScrollToRow !== newScrollToRow) {
```
offirgolan: Can this just be `else if (newScrollToRow instanceof Row && oldScrollToRow !== newScrollToRow)`?

offirgolan: Another option is to have an `assert` in `didReceiveAttrs`

addon/components/lt-body.js
buschtoens: So instead of using `didUpdateAttrs` we should go back to `didReceiveAttrs` and store private copies of the attributes for diffing?

offirgolan: Yeah. I spoke with locks and confirmed that we should not be using the arguments in `didUpdateAttrs`. Sorry for making you revert 😅

buschtoens:
Okay, so I reverted back to using `didReceiveAttrs` and manual diffing. If I'm not mistaken, that should be all. :laughing: 

Now there's only https://github.com/alphasights/ember-scrollable/pull/26 left :)

offirgolan:
@buschtoens looks like tests are failing.

buschtoens:
Sorry for the silence from my side. I was a bit under the weather over the
course of the last days. I'll finish the PR in a few hours when I'm in the
office. :)

We're also eagerly waiting for this feature :D

PR #235, opened 2016-10-23
[FEATURE] Responsive Columns

### LINKED ISSUE #219: Tables responsive?
Can the tables be response for mobile optimized?

offirgolan:
Resolves #219. 

This PR introduces a responsive solution. Using [ember-responsive](https://github.com/freshbooks/ember-responsive), I have introduced two methods to handle responsive behaviors that will hide the appropriate columns. This PR also introduces a new table property `responsiveHiddenColumns` to be able to easily distinguish between hidden columns and columns hidden from this responsive behavior.

### Via column definition

Specify an array of media breakpoints that determine when this column will be shown. If we have the following breakpoints defined in `app/breakpoints.js`:

- mobile
- tablet
- desktop

And we want to show this column only for tablet and desktop media, the following array should be specified:

```js
{
  breakpoints: ['tablet', 'desktop']
}
```

If this property is `null`, `undefined`, or `[]`, then this column will always be shown, regardless of the current media type.

### Via light-table component

Pass a hash to determine the number of columns to show per given breakpoint. If this is specified, it will override any column specific breakpoints. If we have the following breakpoints defined in `app/breakpoints.js`:

- mobile
- tablet
- desktop

The following hash can be passed in:

```hbs
{{#light-table table breakpoints=(hash 
  mobile=2
  tablet=4
) as |t|}}

{{/light-table}}
```

If there is no rule specified for a given breakpoint (i.e. `desktop`), all columns will be shown.

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -26,6 +26,14 @@ export default class Column extends Ember.Object.extend({
   hidden: false,
 
   /**
+   * If true, this column has been hidden due to the responsive behavior
+   * @property responsiveHidden
+   * @type {Boolean}
+   * @default false
+   */
+  responsiveHidden: false,
+
+  /**
    * @property ascending
    * @type {Boolean}
    * @default true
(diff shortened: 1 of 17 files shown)


offirgolan:
@taras I agree but at the same time I believe `ember-responsive` is pretty widely used and a trust worthy addon to depend on. I just dont think we should be spending time to reinvent the wheel here. If there will be much of a demand to decouple, we can cross that bridge then 😸

taras:
I agree too!

jamesdixon:
You guys are killin' it. Nice work!

PR #250, opened 2016-11-01
Document `Row#content`

knownasilya:
API doesn't make it clear how to get the data, so I think this should help.

### DIFF
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -35,6 +35,14 @@ export default class Row extends Ember.ObjectProxy.extend({
    * @type {String | Array}
    */
   classNames: null,
+  
+  /**
+   * Data content for this row.
+   *
+   * @property content
+   * @type {Object}
+   */
+  content: null,
 
   /**
    * Element ID for the corresponding `{{lt-row}}`.


buschtoens:
`Row` is an `ObjectProxy`. This means that every property not defined on the `ObjectProxy` itself is proxied / forwared to the object that is `content`.

In order to get or set properties, you do not need to access `content`. You can just use the `Row` instance. You only need to acces the `content` object when one of its properties is shadowed by the proxy ('selected' for instance) or when passing it to an Ember Data `Model` as a relationship IIRC.

offirgolan:
I think this is ok. Users should be aware that this property exists.

buschtoens:
I think so too, but I was trying to say that a more elaborate description
would be in order. (?)

offirgolan:
@buschtoens I updated the description to the following which is already in master:

```js
  /**
   * Data content for this row. Since this class extends Ember.ObjectProxy,
   * all properties are forwarded to the content. This means that instead of
   * `row.content.foo` you can just do `row.foo`. Please note that methods are
   * not forwarded. You will not be able to do `row.save()`, instead, you would have
   * to do `row.content.save()`.
   *
   * @property content
   * @type {Object}
   */
  content: null
```

buschtoens:
👍

PR #254, opened 2016-11-04
[BUGFIX] repeated scrollToRow

buschtoens:
Currently repeatedly setting the `scrollToRow` property to the same row doesn't work, because the `targetScrollOffset` isn't properly resetted.

This PR fixes that and also removes the unnecessary checks from #241.


offirgolan:
👍 must have been from my refactoring. My mistake 😝

PR #258, opened 2016-11-09
[FEATURE] Draggable Columns

### LINKED ISSUE #46: Reorder columns
Just wondering if there's any interest in draggable column headings for column re-ordering. This is the one feature missing that is stopping us dropping this in.

Has it been looked at previously? Would you consider a PR for it?

offirgolan:
Resolves #46.

Allows reordering of columns via drag n drop.

## Column Properties

```js
/**
 * Whether the column can be reorder via drag and drop.
 *
 * CSS Classes:
 *  - `is-draggable`
 *  - `is-dragging`
 *  - `is-drag-target`
 *    - `drag-left`
 *    - `drag-right`
 *
 * @property draggable
 * @type {Boolean}
 * @default false
 */
draggable: false

/**
 * Whether the column is a valid drop target.
 *
 * @property droppable
 * @type {Boolean}
 * @default true
 */
droppable: true
```

## Header & Footer Actions

```js
/**
 * onColumnDrag action.
 *
 * @event onColumnDrag
 * @param  {Column} column The column that is being dragged
 */
onColumnDrag(column) {}

/**
 * onColumnDrop action.
 *
 * @event onColumnDrop
 * @param  {Column} column The column that was dropped
 * @param  {Boolean} isSuccess The column was successfully dropped and sorted
 */
onColumnDrop(column, isSuccess) {}
```

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -1,7 +1,9 @@
 import Ember from 'ember';
 
 const {
+  guidFor,
   isEmpty,
+  makeArray,
   computed,
   A: emberArray
 } = Ember;
@@ -12,6 +14,11 @@ const {
  */
 export default class Column extends Ember.Object.extend({
   /**
+   * Whether the column can be hidden.
+   *
+   * CSS Classes:
+   *  - `is-hideable`
+   *
    * @property hideable
    * @type {Boolean}
    * @default true
@@ -19,6 +26,11 @@ export default class Column extends Ember.Object.extend({
   hideable: true,
 
   /**
+   * Whether the column can is hidden.
+   *
+   * CSS Classes:
+   *  - `is-hidden`
+   *
    * @property hidden
    * @type {Boolean}
    * @default false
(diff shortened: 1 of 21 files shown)


vasilakisfil:
@offirgolan is there a way we can get the column that received the drop ?

offirgolan:
@vasilakisfil Not sure I follow. If you want to get the change of index, you can do that by getting the original column's position from `onColumnDrag` and the final position from `onColumnDrop`.

vasilakisfil:
@offirgolan I get the same column (the one I drag) on both `onColumnDrag` and `onColumnDrop`. Should I get a different column ?

offirgolan:
Nope it will be the same column. The actions exist so you can check the state of the column before and after it has been repositioned. 

You can do something like:

```js
table.get('columns').indexOf(column);
```

To get the before and after index.

vasilakisfil:
ah ok cool I will try that. I think it would be nice if we got both column on the onColumnDrop event.

PR #265, opened 2016-11-14
[WIP][FEATURE] Passthrough Properties

### LINKED ISSUE #151: Allow use of a model attribute as the row's selected state property
If possible, it would be great to be able to specify the name of an attribute on a row's associated content/model that is used to hold the row's selected state, with the actual `row.selected` property just becoming a computed alias of that. 

Currently, you need to jump through a few hoops, setup observers/ect, if you want to persist row selection state on the row models and keep them in sync (e.g. when models are added to the table and are already in a selected state). 

Additionally, I think it could make sense to expose row selection/deselection actions that get fired passing the array of affected rows to the action handler, instead of only being able to observe the `table.selectedRows` property. This would be useful for executing side effects, like saving selection state changes to the backend, as opposed to trying to use a `table.selectedRows` observer where its harder to determine what changes actually occurred.

buschtoens:
Closes #151.

This is a stab at implementing properties for `Row`s that are optionally passed through to the `content` object.

The API is similar to the `classNameBindings` property of `Ember.Component` with the extra ability to invert properties.

For example:

```js
new Row(content, { passthrough: ['selected', 'expanded:isShowingMore', 'hidden:!visible'] });
```

Relevant docs from the PR:

```js
  /**
   * Properties listed here are passed through to the `content` object instead
   * of getting shadowed by the `ObjectProxy` behavior of `Row`.
   *
   * You can use mappings similar to `classNameBindings` in `Ember.Component`.
   *
   * - `'selected'` - transparently pass through the `selected` property
   * - `'expanded:isShowingMore'` - map the `expanded` property to the
   *   `isShowingMore` property
   * - `'hidden:!visible'` - map the `hidden` property to the `visible` property
   *   and invert the value
   *
   * @property passthrough
   * @type {Array}
   * @default []
   */
  passthrough: computed(() => []),
```

Do you like the way this implemented?

If you're okay with this, I'd make the required changes in `Table.js` to be able to actually configure a table to use passthrough properties and fix the tests.

### DIFF
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -1,6 +1,35 @@
 import Ember from 'ember';
 
-const { computed, guidFor } = Ember;
+const { get, computed, guidFor, isNone, isArray } = Ember;
+
+function buildPassthroughProperty(defaultValue) {
+  return computed('_passthroughMappings', {
+    get(key) {
+      let mappings = this.get('_passthroughMappings');
+
+      if (!mappings || !mappings[key]) {
+        let value = this.get('_passthroughBuffer')[key];
+        return isNone(value) ? defaultValue : value;
+      }
+
+      let { path, isInverted } = mappings[key];
+      let value = this.get(path);
+      return isInverted ? !value : value;
+    },
+    set(key, value) {
+      let mappings = this.get('_passthroughMappings');
+
+      if (!mappings || !mappings[key]) {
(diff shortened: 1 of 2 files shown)


buschtoens:
Whoops. I just realized that this is a total brain fart, because the properties are only bound in one way. Haha.

I'll blame it on the early morning :D

PR #267, opened 2016-11-15
Add failing test for Ember 2.7

simonihmig:
I added this ember-try scenario to document that tests are actually failing for Ember 2.7 and probably also 2.6, since upgrading ember-wormhole to 0.5.1. Encountered this in one of our apps. Same things happens in ember-bootstrap when using the same wormhole version (see https://github.com/kaliber5/ember-bootstrap/pull/160).

Will file an issue with ember-wormhole...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
   - EMBER_TRY_SCENARIO=default
   - EMBER_TRY_SCENARIO=ember-2.3
   - EMBER_TRY_SCENARIO=ember-2.4
+  - EMBER_TRY_SCENARIO=ember-2.7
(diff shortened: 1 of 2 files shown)


simonihmig:
@offirgolan The added test here does fail, but for some very other reason (some failing node dependency). I have no idea why that is. Runing `ember try:one ember-2.7` I see the exception coming from ember-wormhole that I have encountered, see the screenshot:
image

Maybe you can clear the Travis cache and run the build again? I once had some weird Travis issues because of some stale cache...

offirgolan:
@simonihmig I added this in #282 but I disabled it for now until the ember-wormhole issue gets resolved.

PR #290, opened 2016-12-16
[FEATURE] Add selectOnClick option

### LINKED ISSUE #289: Improve the UX for multiselect
Hi,

I believe that multi-select the way it is shown in the docs is not very intuitive for non power users. Few people know about cmd+click or shift+click. In our product we implemented checkboxes:

Beside adding a visual clue that rows are selectable, it offers the standard UI of "select all" that most people are used to. The issue now is that the user will select rows via the checkboxes but as soon as they click (by mistake or not) on anything, the entire selection is lost. It makes the UX very frustrating:

I was wondering if you would consider more granular options for the multiselect. Basically I would like to avoid that clicking actually select or unselect anything. Clicking should not do anything. But I would like to keep cmd+click and shift+click for power users. Is that something you are willing to consider?

offirgolan:
Resolves #289. 

Select a row on click. If this is set to `false` and multiSelect is enabled, using click + `shift`, `cmd`, or `ctrl` will still work as intended, while clicking on the row will not set the row as selected.

/cc @karellm. Can you please check out this branch and test it out in your app?

### DIFF
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -76,6 +76,17 @@ export default Component.extend({
   canSelect: true,
 
   /**
+   * Select a row on click. If this is set to `false` and multiSelect is
+   * enabled, using click + `shift`, `cmd`, or `ctrl` will still work as
+   * intended, while clicking on the row will not set the row as selected.
+   *
+   * @property selectOnClick
+   * @type {Boolean}
+   * @default true
+   */
+  selectOnClick: true,
+
+  /**
    * Allows for expanding row. This will create a new row under the row that was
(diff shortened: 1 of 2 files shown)


karellm:
@offirgolan Thanks for the prompt answer! Testing it out right now.

karellm:
@offirgolan The behaviour looks exactly like what I had in mind. I'm not sure I'm the best person to do it but let me know if you also want a code review! 

Thanks a lot for the great support 👍

offirgolan:
@karellm great! Np at all, I realized the change was pretty minimal so I took a stab at it. I'll make a new minor release with this later today.

Issue #284, opened 2016-12-09
Horizontal Scroll + Ember-Scrollable


offirgolan:
Opening this issue to track the progress of ember-scrollable#29

cc @alexander-alvarez.

KeithClinard:
Are there any updates on this since #321 has been merged?

I would really like to see horizontal scrolling work with a fixed header.

offirgolan:
@KeithClinard this is already supported (see #222). This issue is about using ember-scrollable's virtual scrollbars instead of the current rudimentary solution.

KeithClinard:
As it appears to be unrelated to this issue, I've created another issue for the problem I am having. (#335)

SteelBurgher:
I see support for both horizontal and vertical scrolling was recently merged into ember-scollable. What's the timeline on integrating it with ember-light-table?

aaronlelevier:
Bump

Is horizontal scroll supported?

I tried adding `ember-scrollable` to wrap the `light-table`, but it didn't work. 

```
{{#ember-scrollable}}
  Some long content...
{{/ember-scrollable}}
```

Any help is appreciated. Thanks

alexander-alvarez:
It's not currently. There's a quite a bit of work to be done with freezing rows/columns/etc which makes the horizontal scroll complete. That I know of, it isn't being tackled at the moment

PR #298, opened 2016-12-23
Calculate the total table width more precisely

mkay581:
The `totalWidth` of the table is calculated by the table headers, however its result calculation does not take in account when these table headers have padding and borders.

This changes the code to calculate the total width of the table based on the `<th>` elements vs. strings in columns, allowing use to use the `offsetWidth` property that includes the element's padding and border in its width output.

### DIFF
--- a/addon/components/light-table.js
+++ b/addon/components/light-table.js
@@ -155,57 +155,39 @@ const LightTable = Component.extend({
   visibleColumns: computed.readOnly('table.visibleColumns'),
 
   /**
-   * Calculates the total width of the visible columns via their `width`
-   * propert.
+   * Calculates the total width of the visible column elements via their `width`
+   * property.
    *
    * Returns 0 for the following conditions
    *  - All widths are not set
    *  - Widths are not the same unit
-   *  - Unit cannot be determined
    *
    * @property totalWidth
    * @type {Number}
    * @private
    */
-  totalWidth: computed('visibleColumns.[]', 'visibleColumns.@each.width', function() {
-    let visibleColumns = this.get('visibleColumns');
-    let widths = visibleColumns.getEach('width');
-    let unit = (widths[0] || '').match(/\D+$/);
+  totalWidth: computed('columnElements.[]', function() {
+    let columnElements = this.get('columnElements');
     let totalWidth = 0;
 
-    if (isEmpty(unit)) {
-      return 0;
+    for (let el of columnElements) {
+      totalWidth += el.offsetWidth;
     }
-
-    unit = unit[0];
-
-    /*
-      1. Check if all widths are present
-      2. Check if all widths are the same unit
-     */
-    for (let i = 0; i < widths.length; i++) {
-      let width = widths[i];
-
-      if (isNone(width) || width.indexOf(unit) === -1) {
-        return 0;
-      }
-
-      totalWidth += parseInt(width, 10);
-    }
-
-    return `${totalWidth}${unit}`;
+    return totalWidth;
   }),
 
   style: computed('totalWidth', 'height', function() {
     let totalWidth = this.get('totalWidth');
     let style = this.getProperties(['height']);
 
-    if (totalWidth) {
-      style.width = totalWidth;
-      style.overflowX = 'auto';
-    }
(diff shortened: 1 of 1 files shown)


addon/components/light-table.js
```diff
@@ -219,6 +201,11 @@ const LightTable = Component.extend({
     if (isNone(media)) {
       this.set('responsive', false);
     }
+
+    // wait until after view's element has rendered in order to query the column elements
+    Ember.run.scheduleOnce('afterRender', () => {
```
mkay581: I'm not sure of the best way to wait until the view's `element` is ready to be queried here. Open to ideas.

mkay581:
Note: this is not an issue on the demo page because the demo page forces all table headers to be `border-box` (via bootstrap.css) which ensures that the width of the headers never exceed their specified widths.

offirgolan:
@mkay581 this seems really overcomplicated and not really a route i want to go in. Can't this all be solved by adding the border-box attribute to the table header and footer elements in the addon's css?

mkay581:
> Can't this all be solved by adding the border-box attribute to the table header and footer elements in the addon's css?

Yeah, but for my project, I'm unable to do that because there are other things that depend on the padding/border being separate from the width of the table header cells. Also, wouldn't that mean that the `border-box` would need to be supplied to the table body cells also? Either way, a fix like that would definitely be simpler but would impede overall flexibility with the use of this package. 

I'd rather not have to fork this and diverge down the path you intend for this package--tell you what... let's discuss what route you'd rather go down with this and we can work together to get it the way you want.

offirgolan:
> Also, wouldn't that mean that the border-box would need to be supplied to the table body cells also?

Yup. 

> there are other things that depend on the padding/border being separate from the width of the table header cells

Most frameworks impose `border-box` as a default for table cells due to this sort of issue. What kind of work are you doing that require the padding/border to be separate from the width of the header cells?

mkay581:
> Most frameworks impose `border-box` as a default for table cells due to this sort of issue.

Not sure what you mean by this? Can you explain? This is the only package I've come across that imposes a `border-box` attribute on the table cells. 

> What kind of work are you doing that require the padding/border to be separate from the width of the header cells?

Adding `border-box` attribute to cells, coupled with the package's forcing of fixed width columns, both assume that a developer always knows the maximum amount of content every cell can ever take up. This is fine. But when you want to show padding/border regardless of that cell content, this becomes a problem using `border-box`. Take the following cell for example:

screen shot 2016-12-24 at 4 05 56 pm

I've specified that I want that cell with the icon to be `14px` because I know that any given icon in that cell will never exceed more than 14px. But you can see this is a problem when I want to also control the padding.

offirgolan:
> Not sure what you mean by this? Can you explain? This is the only package I've come across that imposes a border-box attribute on the table cells.

What I meant was that there are many libraries out there that must enforce specific element style rules in order to achieve specific functionalities. This is no different as we do this in a couple of other places to make sure everything works as expected.

The current implementation for calculating the total width of the table is pretty rudimentary and is supposed to be a temporary solution until #284 get resolved.

What is stopping you from adding the extra padding / margin to the total width of the column in the table definition?

mkay581:
> What is stopping you from adding the extra padding / margin to the total width of the column in the table definition?

1. The padding is an `em` unit, not a `px`... so depending on zooming or pixel-ratio, the px is not always a constant value that I can predetermine, but even if it was,
1. I would have to calculate the width, padding, and border total of every cell (now or in the future) every time I'd want to tweak the padding or border width on the table cells

I understand that you're trying to suggest workarounds, but none of them are flexible, practical and/or maintainable. 

I have some time that I can dedicate to helping out the package, which is why I was willing to work with you on implementing a solution that would be more acceptable.

offirgolan:
> The padding needs to be an em unit, not a px... so depending on zooming or pixel-ratio, the result px value is not always a constant value that I can predetermine, but even if it was,

> I would have to calculate the width, padding, and border total of every cell (now or in the future) every time I'd want to tweak the padding or border width on the table cells

Hmmmm I see. Okay I see your point, let me take a close look into this sometime in the next week.

PR #318, opened 2017-01-26
Change <span> to <i> tag for sorting icons

jewls618:
Updated the `<span>` tag to an `<i>` tag in order to allow Semantic UI icons to appear. Font Awesome icons work in either and `<i>` tag or a <span> tag but Semantic UI icons only show up if they are wrapped in an `<i>` tag.


offirgolan:
@jewls618 Thank you for this. I'm gonna need to release this as a minor release which I'll do later this week.

PR #322, opened 2017-01-31
Fix typo for default colspan

knownasilya:


offirgolan:
lol thanks for catching than! 👍

PR #347, opened 2017-02-27
Modular behaviors + Keyboard support

### LINKED ISSUE #236: Row selection has no keyboard support
One idea is to allow `tab` to focus on each row. While focused, an `enter` (or `space`) press would select that row. If the `ctrl` / `cmd` modifier is used when selecting, it would add/remove the row to the existing selection. If `shift` modifier is used, it would select a range of rows.

While a row is selected, the `up` and `down` arrows could move from one row to another (without activating them) -- this is where the `shift` modifier would be useful.

Row states:
- focused (user used `tab` to focus the row); with a mouse, this is achieve via a `click`
- selected (user hit `enter` while row was `focused`; with a mouse, this is achieved a `click`
- highlighted (user uses `up` or `down` to move between rows without focusing the row); with a mouse, this is achieve via `hovering`

I've built something like this before with angular and it wasn't too bad to implement (but I wasn't working with a general purpose table widget so I could use more hacks and dirty tricks that would be acceptable for a large project).

Would PRs implementing something like this be accepted? I don't really have a ton of time to focus on this right now but perhaps with a little guidance on implementation strategy me or someone else could take a stab at this.

plcarmel:
To resolve #236 

Work in progress.


addon/behaviors/behavior.js
```diff
@@ -0,0 +1,14 @@
+import Ember from 'ember';
+
+export default Ember.Object.extend({
+
+  exclusionGroup: null,
+
+  events: null,
+
+  init() {
+    this._super.call(this, ...arguments);
```
offirgolan: `this._super(...arguments)`

addon/behaviors/multi-select.js
```diff
@@ -0,0 +1,52 @@
...
+    let table = ltBody.get('table');
+    let i = table.get('rows').indexOf(row);
+    f.call(this, i, table);
+    this._prevSelectedIndex = i;
+  },
+
+  onSelectRow(ltBody, row) {
+    let isSelected = row.get('selected');
+    if (this.get('requiresKeyboard')) {
+      this._onRowClick(ltBody, row, function(i, table) {
```
offirgolan: `() => {}`

addon/behaviors/behavior.js
offirgolan: Please refactor to use ES6 syntax

addon/behaviors/multi-select.js
offirgolan: Please refactor to use ES6 syntax

addon/behaviors/row-expansion.js
```diff
@@ -0,0 +1,23 @@
...
+
+  // passed in
+  multiRow: true,
+
+  init() {
+    this._super.call(this, ...arguments);
+    this.events.onToggleExpandedRow = [ 'rowClick:_all' ];
+  },
+
+  onToggleExpandedRow(ltBody, row) {
```
offirgolan: Is there a reason you are passing the entire ltBody component here and not just the table instance?

addon/components/lt-body.js
```diff
@@ -350,6 +365,25 @@ export default Component.extend({
     }
   },
 
+  _onFocusedRowChanged: Ember.observer('table.focusIndex', function() {
+    run.schedule('afterRender', null, () => {
+      let row = this.$('tr.has-focus');
```
offirgolan: Would it be easier to use the `scrollToRow` action here?

addon/index.js
```diff
@@ -138,5 +142,15 @@ import Row from './classes/Row';
...
  */
+
 export default Table;
-export { Table, Column, Row };
+
+export {
+  Table,
+  Column,
+  Row,
+  FocusBehavior,
```
offirgolan: I don't think we need to export the behaviors. Who ever needs them can grab them from the behavior dir.

addon/mixins/has-behaviors.js
```diff
@@ -0,0 +1,270 @@
+import Ember from 'ember';
+import withBackingField from 'ember-light-table/utils/with-backing-field';
+import ekbListenerName from 'ember-keyboard/utils/listener-name';
+
+const { A, observer, on } = Ember;
+
+/*
+ * from ember-keyboard
+ */
+const gatherKeys = function gatherKeys(event) {
```
offirgolan: Is there a reason we are grabbing this entire file from ember-keyboard?

offirgolan:
@plcarmel this is pretty damn cool. I've been meaning to do something like this for a while but never got around to it. Thank you for taking the lead on this. I've started with just some general comments/ questions. I still need to do a deep review on the rest of the code once I get a bit more time on my hands. 

I think the API around applying behaviors could be a bit cleaner but we can work on that once we have a fully working solution 😄

offirgolan:
I think it would be also really handy to support `action` in the behavior helpers as well. That would remove the need to use observers to detect if something was selected, deselected, expanded, etc.

addon/behaviors/behavior.js
plcarmel: (Y)

addon/components/lt-body.js
plcarmel: Possibly, I will look into that.

addon/index.js
plcarmel: Ok, no problem.

addon/mixins/has-behaviors.js
plcarmel: The `gatherKeys` function is not part of their public API but I wanted to use the ember-keyboard syntax for row click event names (rowClick:shift, rowClick:shift+ctrl, etc.), to keep things consistent. So I just copied the method that gather the modifiers keys from the event.

plcarmel:
Thank you @offirgolan ! Yes, I agree the behavior API could be cleaner. I look forward to suggestions on how to improve this.

Having behavior helpers supporting action is a good idea !

addon/behaviors/row-expansion.js
plcarmel: I don't like it either. It was simply to make `mixins/has-behaviors` more generic, i.e. it is simply passing `this` as the first parameter. Not a very good reason, I aggree. Maybe we can instruct `mixins/has-behaviors` what to pass as the first argument. What do you think ?

addon/components/lt-body.js
plcarmel: I am unable to make it work with scrollToRow. I will leave it like this until someone figure this out.

addon/mixins/has-behaviors.js
offirgolan: Ah that was my misunderstanding. I thought you took the entire file from ember-keyboard.

offirgolan:
@plcarmel let me know when you're at a stable place and I'll do a deep dive into this PR 😄 

In terms of the API for the behavior helpers. I think you should just have a single helper.

So instead of:

```hbs
{{#t.body
    behaviors=(array 
                  (lt-row-focus) 
                  (lt-row-expansion multiRow=false)
              )
}}
```

You will have something like:

```hbs
{{#t.body
    behaviors=(array 
                  (lt-behavior 'row-focus') 
                  (lt-behavior 'row-expansion' multiRow=false)
              )
}}
```

addon/behaviors/row-expansion.js
offirgolan: I dont think that a behavior should have interaction with the lt-body component. Everything should be done on the table instance which should be then reflected on the required components in the view layer. 

Passing around the component is considered bad practice since it can be destroyed at any time. I would much rather have the table instance be passed around. 

Maybe even just init the behavior with the table instance? That way you only have to do it once. Im not exactly sure of the cleanest route for that though.

plcarmel:
@offirgolan, it though about it an I think it is _not_ a bad idea to pass the `lt-body` component to the behavior functions instead of just passing the table.

**Here's a use case:**
Lets say I want to implement "pg up / pg down" in `behaviors/row-focus`.  Then I need to be able to decide how many rows fit in the component and, for that, I need to have access to the DOM element, directly or indirectly.

And it is true that the component can be destroyed at any time, but we pass the reference to the functions of the behavior instance, _not_ to the behavior instance itself.  A behavior should not save this reference for future usage and we can put emphasis on this in the documentation.

What do you think ?

offirgolan:
@plcarmel I think your case is valid. 

I actually think that this sort of concept is pretty interesting and can have a huge benefit to a lot of apps / addons in the ecosystem. The ability to define tasks that will be evented via actions / keyboard events and are centralized to a single modular location is pretty damn nice to have. This is why I have taken your general idea and created a new addon for it. 

I took it in a bit of a different direction which I think makes the implementation much cleaner and much more modular. I'd love for you to check it out so we can work on it together. You can check out the repo at [offirgolan/ember-evented-tasks](https://github.com/offirgolan/ember-evented-tasks).

plcarmel:
I share your opinion. That was my plan to eventually make an add-on out of but I am happy that you are taking it on yourself to do it.

My only concert with the concept of evented task is that, by making it too general, we loose some of the properties of behaviors (groups of behaviors that are mutually exclusive, behaviors having a higher priority, etc.) Maybe one can be built on top of the other ?

offirgolan:
> Maybe one can be built on top of the other ?

@plcarmel that's the plan. The idea of the evented tasks to to be as general as possible so other apps/addons can build on top of it. I prefer to have this sort of logic as a 3rd party addon so we can thoroughly test it. 

It seems as though you have come across a good amount of obstacles as you mentioned so maybe we can try to general some of those as much as possible and port it over to `ember-evented-tasks`, the rest of the logic can stay with this PR 😄 

Thoughts?

plcarmel:
Sounds good to me :smile:

offirgolan:
Sweet! Feel free to open up PRs / issues.

offirgolan:
@plcarmel I think that ember-evented-tasks is now in a pretty good state. I have added you as a collaborator to the repo so we can work together on the whole prioritization and mutually exclusive stuff. 

Also, I'm not sure if `tasks` is a fitting name.... maybe we should stick to `behaviors`? Anyways, feel free to reach out to me on the ember community slack and we can chat there 😄

plcarmel:
@offirgolan My strategy right now is to end my implementation of spreadsheat-select in this PR before I start to work on / use your addon. This way, we will end up with something that cover more use cases.

offirgolan:
@plcarmel no rush 😄

plcarmel:
@offirgolan, I hope you are doing well.  I updated the demo and the branch of the PR.

https://plcarmel.github.io/ember-light-table/#/rows/spreadsheet

I implemented almost all the features that I wanted. Selection is now much easier with the mouse. It is even possible to resize existing selections using handles.

However, `behaviors/spreadsheet-select` is now way to way too big and should be separated in multiple (sub-?)behaviors that can be composed into different selection behaviors, to suit the needs of the user. I am not sure exactly of what would be the best way of implementing that. I you have an idea, let me know.

Unfortunately, I now have to take a pause of at least a couple of weeks to work on some urgent stuff.  I will let you know when I am ready to give this some more time.

plcarmel:
@offirgolan, I am back. Turns out it took me three month to finish my urgent work :( I will work full time in the next day to try to wrap this up as quickly as possible.

offirgolan:
@plcarmel that's great to hear! Im super excited to see this PR progress 😄

plcarmel:
@alexander-alvarez just made me realize there is a 2.0 branch. I looked at it and jQuery got removed, which is good. However, I rely heavily on it so I'll need to refactor my code. Once this is done, do you think it could be merged in 2.0 ?

alexander-alvarez:
Hey @plcarmel . Thanks so much for all of this work, and coming back to it even after such a large amount of time. I definitely want to include what we can into ELT!

Would it be possible to break apart your work into more concise changes? This way we can incorporate them incrementally into the main branch.
I leave it up to you on how you would want to divy up if appropriate but from my perspectives I see several different things that could be merged in independently
- Adding APIs for behaviors, and introducing the concept to ELT developers -- low risk
- Restructuring the DOM with the lt-frame -- higher risk (table alignment and scroll is such a finicky thing)
- Implementing specific "behaviors" -- low risk

plcarmel:
Working on it.

plcarmel:
Ok, I did an effort to break it into pieces. "Behaviors" API and basic behaviors are now independent of the new structure. However, interesting behaviors like "row-focus" and "spreadsheet" interact with scrolling and I really don't have the energy to retrofit them to the old structure.

@alexander-alvarez, how do you want to proceed from here ?

alexander-alvarez:
Awesome thank you @plcarmel  I'll need to take some time to thoroughly review your work, but I appreciate all the work you've already done to get this here.

Expect clarifying questions, comments about documentation and also doing this without jquery.

I need to do a better job of highlighting this, but in case you don't know, right now we have a [2.x] branch that has some breaking changes in it (including being jquery-less), so we'd introduce those there. For example, we consider major DOM layout / CSS changes breaking changes.

To that point https://github.com/offirgolan/ember-light-table/pull/347/commits/ea427aa69b8008ea632b863cc707b0ad448972fa and https://github.com/offirgolan/ember-light-table/pull/347/commits/50d3fd1d1bc8d7175ee95c94cc97457e59f08e3f can likely make it into our master branch, whereas https://github.com/offirgolan/ember-light-table/pull/347/commits/fda286cb006748546fa1e4431f543fbc7d534fd2 (and anything dependent on it) may be something that's better to target our 2.x branch.


I'll have more for you once I review, but I wanted to give you heads-up in case you wanted to start thinking about these things with me. Thanks again for your hard work!

[2.x]: https://github.com/offirgolan/ember-light-table/tree/2-x

plcarmel:
What can I do to speed thing up, except removing dependency on jquery, which I'll try to do asap ?

rwwagner90:
👋 We just merged the 3-x branch into master. Can you please rebase your PR?

rwwagner90:
@plcarmel Thanks for the PR! We've recently made a lot of significant changes to ember-light-table. Can you please rebase your PR and make sure everything is still working as intended? If this PR is no longer necessary after the new changes, please close it. Thanks!

PR #375, opened 2017-03-27
Make ember-responsive a dependency in all envs

### LINKED ISSUE #367: Upgrade to ember-cli 2.12 gives Error: Attempting to inject an unknown injection: 'service:media
Following attempted upgrade to ember-cli 2.12, ember-light-table causes the following on pages in which it worked with a previous cli version:
```
Error: Attempting to inject an unknown injection: 'service:media'
    at Registry.validateInjections (ember.debug.js:2574)
    at ember.debug.js:1696
    at runInDebug (ember.debug.js:6095)
    at Object.runInDebug (ember.debug.js:17757)
    at instantiate (ember.debug.js:1690)
    at DeprecatedFactoryManager.create (ember.debug.js:1839)
    at Proxy.create (ember.debug.js:1405)
    at CurlyComponentManager.create (ember.debug.js:12905)
    at OpenComponentOpcode.evaluate (ember.debug.js:47779)
    at VM.execute (ember.debug.js:54370)
```

Please remove this if this is already covered in the [Greenkeeper report for 2.12 - issue 364](https://github.com/offirgolan/ember-light-table/issues/364).

steven-ferguson:
Fixes #367

### DIFF
--- a/blueprints/ember-light-table/index.js
+++ b/blueprints/ember-light-table/index.js
@@ -3,8 +3,4 @@ module.exports = {
   description: 'Install Ember Light Table dependencies',
 
   normalizeEntityName: function() {},
-
-  beforeInstall: function() {
-    return this.addAddonToProject('ember-responsive');
-  }
 };
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "ember-one-way-controls": "2.0.0",
     "ember-owner-test-utils": "0.1.1",
     "ember-resolver": "^2.0.3",
-    "ember-responsive": "2.0.2",
     "ember-suave": "4.0.1",
     "ember-truth-helpers": "1.3.0",
     "loader.js": "^4.0.10",
@@ -70,6 +69,7 @@
     "ember-composable-helpers": "^1.1.2",
     "ember-get-config": "0.2.1",
     "ember-in-viewport": "2.1.1",
+    "ember-responsive": "2.0.2",
     "ember-scrollable": "^0.4.4",
     "ember-truth-helpers": "1.2.0",
     "ember-wormhole": "^0.5.1"


jamesdixon:
@steven-ferguson beat me to it :)

jamesdixon:
thanks for the fix!

offirgolan:
This is not needed. The ember light table blueprint installs ember-responsive on the host app. This is why it's not required as a dependency.

steven-ferguson:
It seems like it would be much more straight forward to include it as a dependency. That way users don't have to worry about maintaining compatibility.

offirgolan:
@steven-ferguson the point here is that ember-responsive *must* exist on the host application. Not as a dependency of ELT. The media service as well as the breakpoints file are both needed for everything to function correctly.

alexander-alvarez:
should we update the readme?

offirgolan:
I dont think there should be a need to. Everything should be correctly setup and installed if you run `ember install ember-light-table`.

steven-ferguson:
[From the ember-cli docs](https://ember-cli.com/extending/#addons-depending-on-other-addons), the proper way to do this is by including as a dependency.

alexander-alvarez:
@steven-ferguson what @offirgolan is trying to say is that the consuming host app needs to have `ember-responsive` as a dependency because it is required as a top-level to define breakpoints and be consumed by ember-responsive https://github.com/freshbooks/ember-responsive#getting-started among other things I imagine

steven-ferguson:
Right, which is exactly what including it as a dependency will ensure. [From the docs](https://ember-cli.com/extending/#addons-depending-on-other-addons):

> If your addon depends on another addon, install it as a dependency in your package.json...
> An application that consumes your addon will automatically install and bundle dependencies specified this way.

alexander-alvarez:
this is the expected output.
```diff
dependency-test $ ember install ember-light-table
Installed packages for tooling via npm.
installing ember-light-table
  install addon ember-responsive
Installed packages for tooling via npm.
installing ember-responsive
  create app/breakpoints.js
  create app/initializers/responsive.js
  create tests/helpers/responsive.js
Installed addon package.
Installed addon package.

~/dependency-test $ git diff
diff --git a/package.json b/package.json
index c6f757d..411889d 100644
--- a/package.json
+++ b/package.json
@@ -33,8 +33,10 @@
     "ember-cli-uglify": "^1.2.0",
     "ember-data": "^2.11.0",
     "ember-export-application-global": "^1.0.5",
+    "ember-light-table": "1.8.4",
     "ember-load-initializers": "^0.6.0",
     "ember-resolver": "^2.0.3",
+    "ember-responsive": "2.0.2",
     "ember-source": "^2.11.0",
     "ember-welcome-page": "^2.0.2",
     "loader.js": "^4.0.10"
diff --git a/tests/.jshintrc b/tests/.jshintrc
index d2bd113..7ec445c 100644
--- a/tests/.jshintrc
+++ b/tests/.jshintrc
@@ -1,5 +1,6 @@
 {
   "predef": [
+    "setBreakpoint",
     "document",
     "window",
     "location",
diff --git a/tests/test-helper.js b/tests/test-helper.js
index e6cfb70..b02ae8c 100644
--- a/tests/test-helper.js
+++ b/tests/test-helper.js
@@ -1,4 +1,6 @@
 import resolver from './helpers/resolver';
+import './helpers/responsive';
+
 import {
   setResolver
 } from 'ember-qunit';
```

what you reference above is stating that the addon's dependencies will automatically be bundled, but we need `ember-responsive` to be installed top level, when you install `ember-light-table`

alexander-alvarez:
see https://github.com/alexander-alvarez/dependency-test/commits/master

steven-ferguson:
If you include `ember-responsive` as a dependency for `ember-light-table`, ember-responsive's `media` service will be available in the consuming application. I have created a demonstration [here](https://github.com/steven-ferguson/nested-addons). In my opinion this is a much simpler way to include dependent addons because the user does not have to rerun ember install every time they update your package.

offirgolan:
@steven-ferguson 

> If you include ember-responsive as a dependency for ember-light-table, ember-responsive's media service will be available in the consuming application.

That is the point. Not only that but the generator for ember-responsive must be run on the consuming application so breakpoints file is created correctly. 

This addon has a couple other dependencies but ember-responsive must be installed on the consuming app there currently is no way around that.

alexander-alvarez:
see https://github.com/freshbooks/ember-responsive/issues/84 and https://github.com/freshbooks/ember-responsive/pull/87

PR #390, opened 2017-04-22
Move ember-truth-helpers to dependencies

fsmanuel:
I removed `ember-truth-helpers` from devDependencies and updated to `1.3.0`.


blimmer:
This is showing up on every yarn install because of this problem:

```
warning ember-truth-helpers@1.2.0: The engine "ember-cli-app-version" appears to be invalid.
```

it'd be great to get this resolved

alexander-alvarez:
Thanks for this. We'll update with the fix version, when released

PR #394, opened 2017-04-25
Update ember-truth-helpers dependency.

blimmer:
It was defined in both dev and regular dependencies, but with two different versions.


blimmer:
er, actually #390 does this too, closing.

PR #397, opened 2017-04-28
Fix broken onScroll handler after v1.8.6

eruizdechavez:
After fixing deprecation message for onScroll, a typo was introduced by using `onScrollToY` instead of `onScrollY`. This PR resolves this problem by using the correct event handler and adds a basic test to ensure the event gets fired properly.

### DIFF
--- a/tests/integration/components/light-table-test.js
+++ b/tests/integration/components/light-table-test.js
@@ -189,3 +189,29 @@ test('passed in components can have computed properties', function(assert) {
 
   assert.equal(this.$('.custom-row.is-active').length, 0, 'none of the items are active');
 });
+
+test('onScroll', function(assert) {
+  let done = assert.async();
+  let table = new Table(Columns, createUsers(10));
+
+  this.setProperties({
+    table,
+    currentScrollY: 0,
+    onScroll() {
+      assert.ok(true, 'onScroll worked');
+      done();
+    }
+  });
+
+  this.render(hbs `
+    {{#light-table table height='40vh' as |t|}}
+      {{t.head fixed=true}}
+      {{t.body
+        useVirtualScrollbar=true
+        onScroll=onScroll
+      }}
+    {{/light-table}}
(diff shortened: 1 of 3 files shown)


tests/integration/components/light-table-test.js
```diff
@@ -189,3 +189,29 @@ test('passed in components can have computed properties', function(assert) {
...
+
+test('onScroll', function(assert) {
+  let done = assert.async();
+  let table = new Table(Columns, createUsers(10));
+
+  this.setProperties({
+    table,
+    currentScrollY: 0,
+    onScroll() {
+      assert.ok(true, 'onScroll worked');
```
alexander-alvarez: nit-picky, but do we want to check for `e.target.scrollTop = 300`?

alexander-alvarez:
thanks for this @eruizdechavez 
For posterity this was introduced in https://github.com/offirgolan/ember-light-table/pull/385

tests/integration/components/light-table-test.js
eruizdechavez: The actual value was slightly different, probably because of row heights, table size, etc. Just wanted to check the event handler was properly called on scroll.

Should I improve the test to also check for position or is this test good enough?

alexander-alvarez: it's probably because 300 > than the total row heights combined, did you try with a smaller value (in my experience it matches).
that being said even something as `assert.ok(e.target.scrollTop > 0)` proves your point that it did scroll.
`assert.equal(e.target.scrollTop, scrollTopAmount)` verifies that it scrolled the requested amount as well as it scrolled.

if it's not too hard I would try to add it

eruizdechavez: Done 😃 .

`300` was indeed too much scrolling for this test, `50` does the trick.

alexander-alvarez:
Thanks. Will update when released

PR #408, opened 2017-05-18
Update ember-scrollable version

### LINKED ISSUE #396: Wrong width calculation
Problem: Table is larger than container
Description: I use paginated table and  scrollable container is more larger, columns are wide 
Resolve my problem with downgrade ember-scrollable at 0.4.4
Correct or are there any other problems?

Edit:  the problem is on ember-srollable 0.4.5

RustyToms:
closes #396


buschtoens:
Nice one. Forgot about that. Thanks. 😊

PR #410, opened 2017-05-24
Resolves a bug caused by dragging columns/text onto other light tables

jplamb:
I ran into this error when working with a list of separate ember-light-tables. Highlighting text and dragging it and the column onto another light table throws a `TypeError: Cannot read property 'get' of undefined`. 

My best guess is the event is being handled by a table that doesn't know about the column being dragged. It potentially skips the dragStart function. The column and/or source column are undefined in this case.

### DIFF
--- a/addon/mixins/draggable-column.js
+++ b/addon/mixins/draggable-column.js
@@ -41,6 +41,8 @@ export default Ember.Mixin.create({
 
   isDropTarget: computed(function() {
     let column = this.get('column');
+
+    if (!column || !sourceColumn) { return false; }
     /*
       A column is a valid drop target only if its in the same group
      */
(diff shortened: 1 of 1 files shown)


alexander-alvarez:
@jplamb sorry for the delay... I've been quite busy..
I understand how this could happen, but I was testing out the concept and was unable to reproduce.
https://github.com/jplamb/ember-light-table/compare/draggable-column-undefined...alexander-alvarez:draggable-column-undefined-testing?expand=1

any ideas?

alexander-alvarez:
@jplamb Where are we on this PR? have you continued to see issues?

jplamb:
Haven't had the time to investigate this further. Although I no longer suspect the error had to do with multiple tables.

PR #418, opened 2017-06-01
Ensure the drop target remains properly identified for draggable columns

### LINKED ISSUE #417: Drag target classes are removed when dragging over child elements
[Special classes](http://offirgolan.github.io/ember-light-table/docs/classes/Column.html#property_draggable) are added to the cells that are being dragged over, so that they can be styled appropriately. However, `dragLeave` is not just triggered when the drag exits the cell, but also when the drag enters a child element inside the cell.

If the table is implemented with draggable columns, and also with either child elements inside the table header cells, or resizable columns, the `is-drag-target` class will be removed from the cell even while it is being dragged over. Resizable columns add a extra div (with class `lt-column-resizer`) inside the table header cell. This means that with both draggable and resizable columns, the drag target styling will rarely be applied when dragging to the left, while it mostly works while dragging to the right.

inconsistent_drag_target

Here is an Ember-twiddle demonstrating the problem: https://ember-twiddle.com/4ab83dd0909fe3bc2083a47c1b0ba6fb?openFiles=components.expandable-table.js%2C

RustyToms:
closes #417 

[Special classes](http://offirgolan.github.io/ember-light-table/docs/classes/Column.html#property_draggable) are added to the cells that are being dragged over, so that they can be styled appropriately. However, `dragLeave` is not just triggered when the drag exits the cell, but also when the drag enters a child element inside the cell.

If the table is implemented with draggable columns, and also with either child elements inside the table header cells, or resizable columns, the `is-drag-target` class will be removed from the cell even while it is being dragged over. Resizable columns add a extra div (with class `lt-column-resizer`) inside the table header cell. This means that with both draggable and resizable columns, the drag target styling will rarely be applied when dragging to the left, while it mostly works while dragging to the right.

inconsistent_drag_target

Here is an Ember-twiddle demonstrating the problem: https://ember-twiddle.com/4ab83dd0909fe3bc2083a47c1b0ba6fb?openFiles=components.expandable-table.js%2C

There are a few different ways to approach this problem but there are annoying edge cases that introduce new bugs for a few of them that I tried. This solution is to reset `isDragTarget` to true in `dragOver`, since it is constantly being called while the cell is being dragged over. It seems to work pretty well.

consistent_drag_target
Here is a Twiddle demonstrating this solution: https://ember-twiddle.com/e30d04f1e3e1f4cdb3f95749dbeb3f6c?openFiles=components.expandable-table.js%2C

### DIFF
--- a/addon/mixins/draggable-column.js
+++ b/addon/mixins/draggable-column.js
@@ -84,6 +84,14 @@ export default Ember.Mixin.create({
 
     if (this.get('isDropTarget')) {
       e.preventDefault();
+      /*
+        NOTE: dragLeave will be triggered by any child elements inside the
+        column. This code ensures the column being dragged over continues to be
+        identified as the current drop target
+       */
+      if (!this.get('isDragTarget')) {
+        this.set('isDragTarget', this.get('column') !== sourceColumn);
+      }
     }
   },
 


alexander-alvarez:
@RustyToms thanks for this. Sorry for the delay.
Do you think the draggable portion of the demo would benefit from having some css associated with these styles as you've demonstrated in your gif?

offirgolan:
@alexander-alvarez the demo page already has some [(minimal) CSS](https://github.com/offirgolan/ember-light-table/blob/master/tests/dummy/app/styles/table.less#L33-L41) to show a valid drop target. 

Can you verify this work across grouped columns as well?

Techn1x:
Just encountered this issue today and also made a video.......... but looks like I don't need to post it :)
Glad it's already known about though!

RustyToms:
Hey @offirgolan @alexander-alvarez this is still a pretty bad issue, what do you need right now in order to merge this? Do you still want the demo updated?

alexander-alvarez:
@RustyToms so sorry this slipped through the cracks... I checked out your branch it and clearly works as expected, so I'll merge this in.

I don't use the drag/drop functionality, but if you feel that we could do a better job on the documentation side of it, please let us know, or open up a PR.

Issue #404, opened 2017-05-11
Cannot read property 'getProperties' of null


Gorzas:
I've just upgraded ember-light-table on my project and I've found that the tests of my projected have started to fail. After some searching I've found that the error is thrown at the [base cell component](https://github.com/offirgolan/ember-light-table/blob/v1.8.6/addon/components/cells/base.js#L29-L31):

```
style: computed('column.width', function() {
    return cssStyleify(this.get('column').getProperties(['width']));
  }),
```

This computed property breaks when the attribute column is set to null (as it is by default). It's a minor issue but it breaks the code and I thing that we can avoid these situations very easy. We have two options:
- Setting the attribute `column` to a default `Column` object instead null.
- Checking if `column` exists before resolving the computed property `style`.

What do you think?

offirgolan:
@Gorzas why is the column being set to `null` in the first place?

Gorzas:
@offirgolan The column is set to `null` by default.

I've been thinking that where it fails is in the cell tests, so maybe it would be better to improve the [blueprint](https://github.com/offirgolan/ember-light-table/blob/master/blueprints/cell-type-test/files/tests/integration/components/light-table/cells/__name__-test.js#L10-L13). 

```
test('it renders', function(assert) {
  this.render(hbs`{{light-table/cells/<%= dasherizedModuleName %>}}`);
  assert.equal(this.$().text().trim(), '');
});
```

The `it renders` is going to fail always because it doesn't pass a column to the component. :/

buschtoens:
@Gorzas In the best case scenario `it renders` provides as minimal input to the component as possible. Even though it does not make any sense to invoke a cell without `column` and `row`, it could theoretically happen in some weird edge-case scenario, possibly crashing the whole Glimmer VM.

Therefore, instead of changing the test case, I would rather "fix" the getter in the actual implementation of `BaseCell`.

PR #421, opened 2017-06-08
lt-body: add enableScaffolding option

buschtoens:
This PR adds a scaffolding row (just like with `{{lt-head}}`) to the `{{lt-body}}` component.

This way it is possible to have a first row with colspans.

The scaffolding row can optionally be enabled like so:

```hbs
{{t.body enableScaffolding=true}}
```

Also removes the now unnecessary style attribute binding from `cells/base`, which in theory should yield a small performance boost. :rocket:

### DIFF
--- a/addon/components/cells/base.js
+++ b/addon/components/cells/base.js
@@ -24,10 +24,18 @@ const Cell = Component.extend({
   attributeBindings: ['style'],
   classNameBindings: ['align', 'isSorted', 'column.cellClassNames'],
 
+  enableScaffolding: false,
+
   isSorted: computed.readOnly('column.sorted'),
 
-  style: computed('column.width', function() {
-    return cssStyleify(this.get('column').getProperties(['width']));
+  style: computed('enableScaffolding', 'column.width', function() {
+    let column = this.get('column');
+
+    if (this.get('enableScaffolding') || !column) {
+      return '';
+    }
+
+    return cssStyleify(column.getProperties(['width']));
   }),
 
   align: computed('column.align', function() {
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -163,6 +163,21 @@ export default Component.extend({
   overwrite: false,
 
   /**
+   * If true, the body will prepend an invisible `<tr>` that scaffolds the
+   * widths of the table cells.
+   *
+   * ember-light-table uses [`table-layout: fixed`](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout).
+   * This means, that the widths of the columns are defined by the first row
+   * only. By prepending this scaffolding row, widths of columns only need to
+   * be specified once.
+   *
+   * @property enableScaffolding
+   * @type {Boolean}
+   * @default false
+   */
+  enableScaffolding: false,
+
+  /**
    * ID of main table component. Used to generate divs for ember-wormhole
    *
    * @property tableId
--- a/tests/integration/components/lt-body-test.js
+++ b/tests/integration/components/lt-body-test.js
@@ -191,6 +191,44 @@ test('hidden rows', function(assert) {
   assert.equal(this.$('tbody > tr').length, 4);
 });
 
+test('scaffolding', function(assert) {
+  const users = createUsers(1);
+  this.set('table', new Table(Columns, users));
+
+  this.render(hbs `{{lt-body table=table sharedOptions=sharedOptions enableScaffolding=true}}`);
+
+  const scaffoldingRow = this.$('tr:eq(0)');
+  const userRow = this.$('tr:eq(1)');
+  const userCells = userRow.children('.lt-cell');
+
+  assert.ok(
+    scaffoldingRow.hasClass('lt-scaffolding-row'),
+    'the first row of the <tbody> is a scaffolding row'
+  );
+  assert.ok(
+    !userRow.hasClass('lt-scaffolding-row'),
+    'the second row of the <tbody> is not a scaffolding row'
+  );
+
+  assert.equal(
+    userRow.attr('style'),
+    null,
+    'the second row of the <tbody> has no `style` attribute'
+  );
+
+  assert.ok(
+    Columns
+      .map((c, i) => {
+        const configuredWidth = Number.parseInt(c.width, 10);
+        const actualWidth = userCells.eq(i).width();
+
+        return configuredWidth ? configuredWidth === actualWidth : true;
+      })
+      .every(Boolean),
+    'the first actual data row has the correct widths'
+  );
+});
+
 test('overwrite', function(assert) {
   this.set('table', new Table(Columns, createUsers(5)));
 
--- a/addon/templates/components/lt-body.hbs
+++ b/addon/templates/components/lt-body.hbs
@@ -15,6 +15,14 @@
 
     <table class={{tableClassNames}}>
       <tbody class="lt-body">
+        {{#if enableScaffolding}}
+          <tr class="lt-scaffolding-row">
+            {{#each columns as |column|}}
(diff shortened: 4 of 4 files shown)


buschtoens:
Tests failing... I'll look into that.

buschtoens:
The tests for `{{lt-body}}` are now aware of `.lt-scaffolding-row` and ignore these.

addon/components/cells/base.js
```diff
@@ -21,15 +20,10 @@ const Cell = Component.extend({
   layout,
   tagName: 'td',
   classNames: ['lt-cell'],
-  attributeBindings: ['style'],
   classNameBindings: ['align', 'isSorted', 'column.cellClassNames'],
 
   isSorted: computed.readOnly('column.sorted'),
 
-  style: computed('column.width', function() {
```
alexander-alvarez: Are we concerned that this no longer happens when `enableScaffolding` is set to false?

buschtoens: I would argue no. Currently, these `style` attributes don't really have an effect anyway, as the `table-layout: fixed` makes all but the first row ignore custom widths.

As long as the user doesn't overwrite the [`table-layout: fixed` rule](https://github.com/offirgolan/ember-light-table/blob/b0db5b15d96e1b7bfd8f96a3001bfac6338ec0b3/addon/styles/addon.css#L14), nothing should change.

The only case where this could matter is, when the user actively decides to change the [`display`](https://developer.mozilla.org/en-US/docs/Web/CSS/display?v=control) from `table` to something like `flex`.

buschtoens: I can undo this, or make it depend on whether `enableScaffolding` is `true` or `false`.

addon/components/lt-body.js
```diff
@@ -163,6 +163,21 @@ export default Component.extend({
...
+   * ember-light-table uses [`table-layout: fixed`](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout).
+   * This means, that the widths of the columns are defined be the first row
+   * only. By prepending this scaffolding row, widths of columns only need to
+   * be specified once.
+   *
+   * @property enableScaffolding
+   * @type {Boolean}
+   * @default true
+   */
+  enableScaffolding: true,
```
buschtoens: Maybe it would be more sensible to make this a CP that's the inverse of [`overwrite`](https://github.com/buschtoens/ember-light-table/blob/fa85616464e1fdbd6b27d1c2e9ef8746d1f833dc/addon/components/lt-body.js#L155-L163).

As we're prepending a scaffolding row, this could lead to some unexpected behavior for people that use selectors like `:first-child`.

buschtoens:
@offirgolan Would you like to chime in on the issues raised above? I'd like move this PR forward. :)

offirgolan:
@buschtoens this is a risky change as it could introduce many unanticipated issues. 

Can you give me an example as to how you would create a table with the first row having colspans? Maybe we can tackle this in a different way.

buschtoens:
I needed implement headings for table sub sections, basically a slimmed down version of the headings in #434 (sticky rows). 

I tested my branch in our project and couldn't find any problems, but i can understand the hesitation and kinda feel the same. Maybe we could inverse the `enableScaffolding` so users explicitly have to opt-in? Just like with `enableSync`. 

Alternatively, I'm sure I'll be able to come up with some `position: absolute` magic to achieve faux colspan.

buschtoens:
I made `enableScaffolding` default to `false` and re-added the `style` binding on `BaseCell`. This way, if users don't explicitly opt-in, nothing changes at all. I also fixed #404 while I was at it.

When `enableScaffolding=true` the `style` CP short circuits and always returns `null`. This way we save some unnecessary DOM updates.

tests/integration/components/lt-body-test.js
```diff
@@ -33,7 +33,7 @@ test('row selection - enable or disable', function(assert) {
 
   this.render(hbs `{{lt-body table=table sharedOptions=sharedOptions canSelect=canSelect}}`);
 
-  let row = this.$('tr:first');
+  let row = this.$('tr:not(.lt-scaffolding-row):first');
```
alexander-alvarez: out of curiousity did you need to update these tests? `enableScaffolding` isn't set

buschtoens: In the original state of this PR with `enableScaffolding` defaulting to `true`, this was necessary. It isn't *anymore*, but I precautiously still included this commit. This way the tests will continue to work, even if we decide to set `enableScaffolding` by default in the future, or when we copy and paste these cases to write some new `enableScaffolding` specific tests.

If you feel that this makes the tests harder to comprehend / maintain, I have absolutely not objections to cancel out this commit. :slightly_smiling_face:

alexander-alvarez: Got it. That makes sense.

It's a little confusing since it isn't set.

As an aside there aren't any actual tests that verify `enableScaffoliding` puts a scaffolding row in the DOM :astonished:  :rotating_light: :roll_eyes:

buschtoens: That is... correct. 😨 

I could swear, that I've written these. 🤔  
Gonna push the tomorrow from the office. ☺️

buschtoens:
I added a test case asserting, that:

- the first row of the <tbody> is a scaffolding row
- the second row of the <tbody> is not a scaffolding row
- the second row of the <tbody> has no `style` attribute
- the first actual data row has the correct widths

tests/integration/components/lt-body-test.js
```diff
@@ -191,6 +191,44 @@ test('hidden rows', function(assert) {
...
+
+  const scaffoldingRow = this.$('tr:eq(0)');
+  const userRow = this.$('tr:eq(1)');
+  const userCells = userRow.children('.lt-cell');
+
+  assert.ok(
+    scaffoldingRow.hasClass('lt-scaffolding-row'),
+    'the first row of the <tbody> is a scaffolding row'
+  );
+  assert.ok(
```
alexander-alvarez: `assert.notOk`

buschtoens: Actually I was gonna use that, but I wanted to stay in line with the other tests in this file, which all use

```js
assert.ok(!condition);
```

Should I leave it or still change to `notOk`?

alexander-alvarez: lol I didn't even realize.
leave it for now... I'll do a PR to migrate to using `ember-native-dom-helpers` in the tests and I'll fix it up there

buschtoens:
@offirgolan Any objections? `enableScaffolding` now defaults to `false`, so users have to opt-in explicitly. There are also sufficient test cases asserting, that the actual data rows have the specified widths. 

Unless you don't want this feature in ELT, I'll merge. 😉

addon/components/cells/base.js
```diff
@@ -24,10 +24,21 @@ const Cell = Component.extend({
...
-    return cssStyleify(this.get('column').getProperties(['width']));
+  style: computed('enableScaffolding', 'column.width', function() {
+    if (this.get('enableScaffolding')) {
+      return null;
+    }
+
+    let column = this.get('column');
+    if (!column) {
+      return null;
+    }
```
offirgolan:
```js
let column = this.get('column');

if (this.get('enableScaffolding') || !column) {
  return '';
}
```

tests/integration/components/lt-body-test.js
offirgolan: @buschtoens I dont think this is needed since you have specific tests for when it is enabled. I prefer it like that since it is more contained. If we enabled it by default in the future, we can just turn it off for these test cases.

buschtoens:
Thanks for reviewing guys. I made the requested changes. Tests are passing as well. :shipit:

PR #426, opened 2017-06-14
Removing setting of a row's elementId

alexander-alvarez:
Related to https://github.com/html-next/vertical-collection/issues/95


buschtoens:
Browser support table for reference: http://caniuse.com/#feat=queryselector

alexander-alvarez:
thanks @buschtoens. Because of this, I'm actually going to change to use https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

Issue #430, opened 2017-06-15
[META] Modernize ELT -- Browser Compatibility / Ember CLI 2.13


alexander-alvarez:
This issue has been opened for a while https://github.com/offirgolan/ember-light-table/issues/145 and hasn't been addressed.

Since the time that it has been open things have been added to the ecosystem to make the ergonomics of compatibility support better. Specifically [targets](https://www.emberjs.com/blog/2017/04/29/ember-2-13-released.html#toc_targets).

Issues such as https://github.com/offirgolan/ember-light-table/issues/427 and https://github.com/offirgolan/ember-light-table/issues/343 will go away automagically because of the functionality in [babel-preset-env](https://github.com/babel/babel-preset-env/blob/master/data/built-in-features.js#L143). Similarly we'll be more aligned with the community best practices.

### :skull_and_crossbones:  `bower.json`
- [ ] Update `ember-cli` to 2.13, reconfigure to use npm & `ember-source` instead of bower
- [ ] `Faker` / `Pretender` shouldn't be needed in bower.json as those are brought in by `ember-cli-mirage` https://github.com/offirgolan/ember-light-table/blob/master/bower.json
- [ ] `Font-awesome` should be a `devDependency` when we move it to the `package.json` -- from what I could see in this [search](https://github.com/offirgolan/ember-light-table/search?p=1&q=fa&type=&utf8=%E2%9C%93). It's css should be brought in for the demo app by using `ember-cli-build.js`. We should document this, so it's no surprise that the ascending and descending icons don't work if users drag and drop our code snippets, don't have font-awesome, and haven't copied our setup.
- [ ] Kill bower.json

### Compatibility Related
- [ ] Document what features we use that we know aren't the most browser compatible including (https://github.com/offirgolan/ember-light-table/issues/427, https://github.com/offirgolan/ember-light-table/issues/343, and https://github.com/offirgolan/ember-light-table/pull/426) and push them to use targets to replace them

### Other
- [ ] Upgrade from jshint to eslint (enabled once we upgrade ember-cli)

--- 

As always, any and all help is welcome. Feel free to chime in the comments to pick up a task, and I'll add your name alongside the task

---
Reference
http://rwjblue.com/2017/04/21/ember-cli-targets/
[ember-source](https://www.emberjs.com/blog/2017/01/23/ember-2-11-released.html#toc_ember-no-longer-supplied-via-code-bower-code)

Eslint
[...]

Issue #427, opened 2017-06-14
Demo doesn't work in IE11


mastastealth:
Just go to: http://offirgolan.github.io/ember-light-table/ in IE11 and you'll get...
![image](https://user-images.githubusercontent.com/1221228/27154931-d7cc7cee-5124-11e7-88c3-075ff5c5dd5d.png)

buschtoens:
https://github.com/offirgolan/ember-light-table/issues/343#issuecomment-281059309

PR #438, opened 2017-06-21
Modernize ELT, kill bower and enable yarn

buschtoens:
Adressing #430.

### :skull_and_crossbones:  `bower.json`
- [x] 533f6df: Update `ember-cli` to 2.13, reconfigure to use npm & `ember-source` instead of bower
- [x] f964cb0: `Faker` / `Pretender` shouldn't be needed in bower.json as those are brought in by `ember-cli-mirage` https://github.com/offirgolan/ember-light-table/blob/master/bower.json
- [x] 4761317: `font-awesome` should be a `devDependency` when we move it to the `package.json`
- [x] a9154f2: Kill `bower.json`

### Also
- [x] 533f6df: Enable `yarn`
- [x] 84abf5b: Remove `jquery@2` from `bower.json`
- [x] 275b2a9: Upgrade demo to use `bootstrap@3.3.7`
- [x] 533f6df7: Switch from `jshint` to `eslint`
- [x] 73490a9: document `font-awesome` usage, so it's no surprise that the ascending and descending icons don't work when users drag and drop our code snippets, if they don't have `font-awesome`, and haven't copied our setup.
- [x] fd32329c: replace `ember-suave` with the spiritual successor [`eslint-plugin-ember-suave`](https://github.com/DockYard/eslint-plugin-ember-suave)
  - [x] 3d47c1c: configure sensibly and fix new linter errors
- [x] 47d98d9: circumvent https://github.com/babel/babel/issues/5862
- [x] 6d38976: Update all other dependencies

### Likely out of scope
- Properly set up `targets` or `includePolyfill`, so that #427 gets fixed. Was already broken prior to this PR.


tests/dummy/app/templates/components/cookbook/horizontal-scrolling-table.hbs
```diff
@@ -1,6 +1,11 @@
 {{!-- BEGIN-SNIPPET horizontal-scrolling-table --}}
 {{#light-table table height='65vh' as |t|}}
 
+  {{!--
+    In order for `fa-sort-asc` and `fa-sort-desc` icons to work,
+    you need to have ember-font-awesome installed.
```
alexander-alvarez: technically we only need `font-awesome` (e.g. via a CDN), do we care to highlight the difference?

buschtoens: Oh, yeah. Good catch.

```hbs
  {{!--
    In order for `fa-sort-asc` and `fa-sort-desc` icons to work,
    you need to have ember-font-awesome installed or manually include
    the font-awesome assets, e.g. via a CDN.
  --}}
```

Would that be better?

addon/mixins/table-header.js
```diff
@@ -66,13 +67,22 @@ export default Ember.Mixin.create({
   resizeOnDrag: false,
 
   /**
+   * CSS classes to be applied to an `<i class="lt-sort-icon></i>` tag that is
+   * inserted into the column's `<th>` element.
+   *
+   * For instance, if you have installed `ember-font-awesome`, you can set
+   * `iconAscending` to `'fa fa-sort-asc'`, which would yield this markup:
+   * `<i class="lt-sort-icon fa fa-sort-asc"></i>`
+   *
```
buschtoens: In accordance with https://github.com/offirgolan/ember-light-table/pull/438#discussion_r123255492 this might be better as well?

```js
  /**
   * CSS classes to be applied to an `<i class="lt-sort-icon></i>` tag that is
   * inserted into the column's `<th>` element.
   *
   * For instance, if you have installed `ember-font-awesome` or include the
   * `font-awesome` assets manually (e.g. via a CDN), you can set
   * `iconAscending` to `'fa fa-sort-asc'`, which would yield this markup:
   * `<i class="lt-sort-icon fa fa-sort-asc"></i>`
   *
   * @property iconAscending
   * @type {String}
   * @default ''
   */
```

alexander-alvarez: in lieu of having documentation that explains what icon fonts are & how they work (I couldn't find anything on the web), this well explained documentation with an example will do :+1:

buschtoens:
Thanks for reviewing! :smiley:

PR #439, opened 2017-06-22
Polyfill support for __proto__ in IE <= 10

### LINKED ISSUE #436: [IE <= 10] Column, Row, Table: static properties (reopen, ...) not inherited
I have a component using the table that does a little bit of customizations via reopen. Oddly enough, only 2 of the 10 seem to fail. I get a:
`TypeError: Object doesn't support property or method 'reopen'` 
...in IE10 for `TableHeaderMixin` and `ColumnClass` (though disabling `ColumnClass` makes the other work, so I think it is the maint culprit). 

However, `reopen`ing `LightTable`, `Header`, `Footer`, `Row`, `Column`, `Cell`, and `SortableObjects` works just fine. Firefox reopens all of them too, so is there anything in particular I should know about `ColumnClass` giving errors? 🤔 

As an example, when I have IE10 console log the `ColumnClass` for me (before reopening) it returns:
```
function Column() {
      var options = arguments.length <= 0 || arguments[0] === undefined ? {} : arguments[0];

      _classCallCheck(this, Column);

      if (options instanceof Column) {
        return options;
      }

      _get(Object.getPrototypeOf(Column.prototype), 'constructor', this).call(this);
      this.setProperties(options);

      var subColumns = options.subColumns;

      subColumns = emberArray(makeArray(subColumns).map(function (sc) {
        return new Column(sc);
      }));
      subColumns.setEach('parent', this);

      this.set('subColumns', subColumns);
    } 
```

which seems to be the correct function. Unfortunately I tried to get a twiddle of this, but ember-twiddle doesn't work on IE10 either! 😅

buschtoens:
Fixes #436.

This gives full support for stuff like this in IE <= 10:

```js
Table.reopen({
  getFirstColumn() {
    return this.get('columns.firstObject');
  }
});
```

---

### Juicy technical details

[MDN on `Object.prototype.__proto__`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/proto)

> ### [Internet Explorer](https://babeljs.io/docs/usage/caveats/#internet-explorer-classes-10-and-below-)
> #### Classes (10 and below)
> If you’re inheriting from a class then static properties are inherited from it via `__proto__`, this is widely supported but you may run into problems with much older browsers.
>
> **NOTE:** `__proto__` is not supported on IE <= 10 so static properties will not be inherited. See the [protoToAssign](https://babeljs.io/docs/plugins/transform-proto-to-assign/) for a possible work around.

> # [babel-plugin-transform-proto-to-assign](https://github.com/babel/babel/tree/64eafad472ebac6333671fff65a9669739e6cd88/packages/babel-plugin-transform-proto-to-assign)
>
> > This plugin allows Babel to transform all `__proto__` assignments to a method that will do a shallow copy of all properties.
>
> ## Detail
>
> This means that the following **will** work:
>
> ```javascript
> var foo = { a: 1 };
> var bar = { b: 2 };
> bar.__proto__ = foo;
> bar.a; // 1
> bar.b; // 2
> ```
>
> however the following **will not**:
>
> ```javascript
> var foo = { a: 1 };
> var bar = { b: 2 };
> bar.__proto__ = foo;
> bar.a; // 1
> foo.a = 2;
> bar.a; // 1 - should be 2 but remember that nothing is bound and it's a straight copy
> ```
>
> This is a case that you have to be aware of if you intend to use this plugin.

### DIFF
--- a/tests/unit/classes/column-test.js
+++ b/tests/unit/classes/column-test.js
@@ -29,6 +29,11 @@ test('create column - column instance', function(assert) {
   assert.equal(col2.label, 'Name');
 });
 
+test('reopen colum', function(assert) {
+  assert.equal(typeof Column.reopen, 'function', 'reopen is a function');
+  assert.equal(typeof Column.reopenClass, 'function', 'reopenClass is a function');
+});
+
 test('CP - isGroupColumn', function(assert) {
   let col = new Column();
   assert.ok(col);
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import fixProto from 'ember-light-table/utils/fix-proto';
 
 const {
   guidFor,
@@ -344,3 +345,6 @@ export default class Column extends EmberObject.extend({
     this.set('subColumns', subColumns);
   }
 }
+
+// https://github.com/offirgolan/ember-light-table/issues/436#issuecomment-310138868
+fixProto(Column);
--- a/addon/classes/Row.js
+++ b/addon/classes/Row.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import fixProto from 'ember-light-table/utils/fix-proto';
 
 const {
   computed,
@@ -100,3 +101,6 @@ export default class Row extends ObjectProxy.extend({
     this.set('content', content);
   }
 }
+
+// https://github.com/offirgolan/ember-light-table/issues/436#issuecomment-310138868
+fixProto(Row);
--- a/addon/classes/Table.js
+++ b/addon/classes/Table.js
@@ -3,6 +3,7 @@ import Row from 'ember-light-table/classes/Row';
 import Column from 'ember-light-table/classes/Column';
 import SyncArrayProxy from 'ember-light-table/-private/sync-array-proxy';
 import { mergeOptionsWithGlobals } from 'ember-light-table/-private/global-options';
+import fixProto from 'ember-light-table/utils/fix-proto';
 
 const {
   get,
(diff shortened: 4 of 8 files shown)


tests/unit/classes/column-test.js
```diff
@@ -29,6 +29,23 @@ test('create column - column instance', function(assert) {
...
+
+  Column.reopen({
+    someMethod() {
+      return true;
+    }
+  });
+  let col = new Column();
+  assert.ok(col.someMethod());
+});
+
```
buschtoens: Not really sure, if these tests make a lot of sense, since we don't run CI tests for IE anyways.

Also, they mutate the Classes. This means that all following test cases have access to `someStaticMethod` and `someMethod`. This might not be a problem now (and I doubt it ever will), but it could and feels kind of dirty.

buschtoens: I think, I should rather write a specialized unit test for `utils/fix-proto`. This way, there's no mutation of global classes.

buschtoens:
There's now a dedicated test for `fixProto`, asserting that ES6 classes that extend `Ember.Object` can be reopened after `fixProto` was used on them.

The testes for `Column`, `Row` and `Table` now merely assert that `reopen` and `reopenClass` are functions.

All these tests are pretty meaningless though, as we don't test in IE <= 10. But now we *could* verify regressions quickly, if someone reports this issue again. :man_shrugging:

offirgolan:
@buschtoens I think we should just move to pure Ember.Object in the next major release because this is kinda getting gross lol.

PR #449, opened 2017-07-05
head & foot: make `table.height` optional, warn in `table-header` mixin

### LINKED ISSUE #446: Don't require `height` property for fixed header / footer
When setting `{{t.head fixed=true}}` or `{{t.foot fixed=true}}` the table body is wrapped inside an `{{ember-scrollable}}` container. Because of that, the height of the table cannot be determined and must be specified explicitly.

Currently, this is done by setting the `height` attribute, like `{{table height="65vh"}}`.  The rendered wrapping `<div>` then looks like this:

```html
<div style="height: 65vh" class="ember-light-table ember-view">
  <!-- ... -->
</div>
```

If no `height` attribute was passed, an error is thrown and the render crashes.

I often set table heights via CSS or even use flexbox layouts. Being forced to set `height` properties when using a fixed header / footer is a nuisance.

I understand that these assertions were put there with good intent, but IMHO they cause more harm than good.

---

[`lt-head.js`](https://github.com/offirgolan/ember-light-table/blob/fcef602bfd2101ee75586d5dd005df9bc1c9841f/addon/components/lt-head.js#L54)
```js
  init() {
    this._super(...arguments);

    let sharedOptions = this.get('sharedOptions') || {};
    let fixed = this.get('fixed');

    assert('[ember-light-table] The height property is required for fixed header', !fixed || fixed && !isEmpty(sharedOptions.height));

    set(sharedOptions, 'fixedHeader', fixed);
  }
```

[`lt-foot.js`](https://github.com/offirgolan/ember-light-table/blob/fcef602bfd2101ee75586d5dd005df9bc1c9841f/addon/components/lt-foot.js#L52)
```js
  init() {
    this._super(...arguments);

    let sharedOptions = this.get('sharedOptions') || {};
    let fixed = this.get('fixed');

    assert('[ember-light-table] The height property is required for fixed footer', !fixed || fixed && !isEmpty(sharedOptions.height));

    set(sharedOptions, 'fixedFooter', fixed);
  }
```

buschtoens:
Fixes #446.

### DIFF
--- a/addon/mixins/table-header.js
+++ b/addon/mixins/table-header.js
@@ -2,7 +2,9 @@ import Ember from 'ember';
 
 const {
   computed,
-  Mixin
+  isEmpty,
+  Mixin,
+  warn
 } = Ember;
 
 /**
@@ -104,6 +106,19 @@ export default Mixin.create({
   sortIcons: computed('iconAscending', 'iconDescending', function() {
     return this.getProperties(['iconAscending', 'iconDescending']);
   }).readOnly(),
+
+  init() {
+    this._super(...arguments);
+
+    let fixed = this.get('fixed');
+    let height = this.get('sharedOptions.height');
+
+    warn(
+      'You did not set a `height` attribute for your table, but marked a header or footer to be fixed. This means that you have to set the table height via CSS. For more information please refer to:  https://github.com/offirgolan/ember-light-table/issues/446',
+      !fixed || fixed && !isEmpty(height),
+      { id: 'ember-light-table.height-attribute' }
+    );
+  },
 
   actions: {
     /**
--- a/addon/components/lt-foot.js
+++ b/addon/components/lt-foot.js
@@ -3,10 +3,9 @@ import layout from 'ember-light-table/templates/components/lt-foot';
 import TableHeaderMixin from 'ember-light-table/mixins/table-header';
 
 const {
-  assert,
   Component,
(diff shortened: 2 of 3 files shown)


buschtoens:
When no `height` attribute is specified, only a warning is printed:

> WARNING: You did not set a `height` attribute for your table, but marked a header or footer to be fixed. This means that you have to set the table height via CSS. For more information please refer to:  https://github.com/offirgolan/ember-light-table/issues/446

addon/mixins/table-header.js
```diff
@@ -105,6 +107,19 @@ export default Mixin.create({
     return this.getProperties(['iconAscending', 'iconDescending']);
   }).readOnly(),
 
+  init() {
+    this._super(...arguments);
+
+    let fixed = this.get('fixed');
+    let height = this.get('sharedOptions.height');
+
+    warn(
```
alexander-alvarez: should we wrap in a https://www.emberjs.com/api/#method_runInDebug

buschtoens: Shouldn't be necessary as `warn` is supposed to be an empty function in a production build. I have not actually verified this though.

https://www.emberjs.com/api/#method_warn

buschtoens: Just checked `Ember.warn` in our production deployment. It's indeed an empty function. :smile:

PR #451, opened 2017-07-05
Table: add `setRowsSynced` method

### LINKED ISSUE #414: enableSync support for setRows
Currently it's not possible to replace a table's synced array with a different synced array.

I suggest that when `{ enableSync: true }` is passed to the `Table` constructor, `setRows` should create a new `RowSyncArrayProxy`.

buschtoens:
Fixes #414.

The tests assert that:

- the table is initialized with a synced array
- the synced array may be replaced
- the replacement array is correctly synced
- mutating the replaced array has no effect

### DIFF
--- a/addon/classes/Table.js
+++ b/addon/classes/Table.js
@@ -192,6 +192,26 @@ export default class Table extends EmberObject.extend({
   }
 
   /**
+   * The same as `setRows`, however the given array is synced, meaning that
+   * mutating the array also updates the table and vice-versa.
+   *
+   * Also see `enableSync` in the constructor options.
+   *
+   * @method setRowsSynced
+   * @param  {Array} rows
+   * @param  {Object} options
+   * @return {Array} rows
+   */
+  setRowsSynced(rows = [], options = {}) {
+    let _rows = RowSyncArrayProxy.create({
+      syncArray: rows,
+      content: emberArray(Table.createRows(rows, options))
+    });
+
+    return this.set('rows', _rows);
+  }
+
+  /**
    * Push the object onto the end of the row array if it is not already present.
    * @method addRow
    * @param  {Object} row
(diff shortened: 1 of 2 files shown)


buschtoens:
> While we're at it: do the row mutation methods (`addRow`, `insertRowAt`, `removeRow`, ...) all work as intended when `{ enableSync: true }` or (if not) are they intended to do so? Should they delegate to the proxied array or throw?

– https://github.com/offirgolan/ember-light-table/issues/414#issuecomment-306396016

Fun fact: they delegate to the synced array. It may be noteworthy that `setRows` *does not* destroy or replace the binding. Instead the synced array's elements are replaced with the given rows.

addon/classes/Table.js
```diff
@@ -192,6 +192,24 @@ export default class Table extends EmberObject.extend({
...
+   *
+   * @method setRows
+   * @param  {Array} rows
+   * @param  {Object} options
+   * @return {Array} rows
+   */
+  setRowsSynced(rows = [], options = {}) {
+    let _rows = RowSyncArrayProxy.create({
+      syncArray: rows,
+      content: emberArray(Table.createRows(rows, options))
```
alexander-alvarez: slightly tangential question: you wouldn't happen to know if there's any downside to https://github.com/buschtoens/ember-light-table/blob/master/addon/classes/Table.js#L155 doesn't pass in the options to `Table.createRows()`

buschtoens: Shoot, I forgot to update the `@method` tag. 

Good question... We could add `options.rowOptions` in the constructor and pass that. If users require options to be passed *right now*, they should be able to wrap in `Table.createRows(rows, options)` methinks.

buschtoens: Shoot, I forgot to update the `@method` tag. 

Good question... We could add `options.rowOptions` in the constructor and pass that. If users require options to be passed *right now*, they should be able to wrap in `Table.createRows(rows, options)` methinks.

PR #457, opened 2017-07-18
light-table: add `extra` property

### LINKED ISSUE #453: Need `tableData` along with `tableActions` to pass down data to custom components
Hi!

I am building out a custom component to use to display my column headers, so that I can filter table results by each column with an additional `{{input}}`.  I am able to successfully pass down an action to handle filtering on `key-up` for my filter input by doing:

```
{{#light-table table tableActions=(hash filterColumn=(perform filterColumn)) as |t|}}
```

I am using the `tableActions` approach, instead of the "curried components" approach laid out in [issue 405](https://github.com/offirgolan/ember-light-table/issues/405).  So I am also passing the initial filter values and state through `tableActions` as a workaround to have that information available to use in my custom column component (say that fast 5 times...).  It may be helpful in the future to split it out into actions and data so it could be done like this:

```
{{#light-table table 
    tableActions=(hash filterColumn=(perform filterColumn)) 
    tableData=(hash filters=filters) as |t|}}
```

Thanks for all your hard work!

buschtoens:
Closes #453

### DIFF
--- a/addon/components/cells/base.js
+++ b/addon/components/cells/base.js
@@ -67,6 +67,12 @@ const Cell = Component.extend({
   tableActions: null,
 
   /**
+   * @property extra
+   * @type {Object}
+   */
+  extra: null,
+
+  /**
    * @property rawValue
    * @type {Mixed}
    */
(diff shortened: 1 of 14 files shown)


buschtoens:
Ya. I should add some for `tableActions` as well then.

PR #458, opened 2017-07-19
Accept an array as `valuePath`

### LINKED ISSUE #456: Accept an array as `valuePath`
I often find myself needing to compute a cell's value dependent of more than one field of the row. But the current API allows only passing one key as the `valuePath`.

I work around this by just accessing whatever properties I need from the `row` attribute that is passed to the `cellComponent`. I think that's pretty ugly.

If we would allow users to pass a string or an array of strings to `valuePath`, then (in the latter case) we could pass an array of values as `value` to the `cellComponent`.

buschtoens:
Fixes #456.

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -239,8 +239,15 @@ export default class Column extends EmberObject.extend({
   cellComponent: null,
 
   /**
+   * The path to the `value` for this column and row. Light Table will basically
+   * call `Ember.get(row, valuePath)` and pass that as `rawValue` to the cell.
+   * The cell then formats `rawValue` and makes it available as `value`.
+   *
+   * If you pass an array, the cell will also receive an array of values as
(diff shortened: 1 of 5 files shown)


buschtoens:
Compared to the rest of `light-table` the performance hit introduced by using `{{collect}}` isn't extreme. Looking at the render performance view in the Ember Dev Tools, I see no noticeable increase in render time for the demo page.

Still, I would like to have some proper metrics first.

danilovaz:
Hi @buschtoens ! Anything with this PR? There is something I can help? The addon is amazing and this feature will help a lot

buschtoens:
@danilovaz glad you're loving it ♥️

I'm hesitant to merge, this adds another dependency and slight performance hit for many users, even though they might never use the feature. 

However, we got you covered. 😉
https://github.com/buschtoens/ember-light-table-cell-type-multi-value

PR #467, opened 2017-07-21
Prettier

buschtoens:
This PR adds the [Prettier](https://prettier.io/) code auto formatter as a fixable eslint rule.

This means that a uniform coding style will be enforced by linter rules. If you violate the coding style, ember-cli will scream warnings. You can also configure your editor to fix eslint errors on save, which I highly recommend. Doing so will ensure that your code is automatically formatted.

I disabled Prettier for some bits, as that could cause pain when upgrading ember-cli.

We adopted Prettier at work and I think it greatly improved our developer experience. 99 % the time it generates easier to read code. For the other 1% there's [`// prettier-ignore`](https://prettier.io/docs/en/usage.html#excluding-code-from-formatting) or `/* eslint-disable prettier/prettier */`.

I have no hard feelings, if you're against merging.

### DIFF
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -2,12 +2,7 @@ import Ember from 'ember';
 import layout from 'ember-light-table/templates/components/lt-body';
 import Row from 'ember-light-table/classes/Row';
 
-const {
-  Component,
-  computed,
-  run,
-  observer
-} = Ember;
+const { Component, computed, run, observer } = Ember;
 
 /**
  * @module Light Table
@@ -327,7 +322,11 @@ export default Component.extend({
   },
 
   onRowsChange: observer('rows.[]', function() {
-    this._checkTargetOffsetTimer = run.scheduleOnce('afterRender', this, this.checkTargetScrollOffset);
+    this._checkTargetOffsetTimer = run.scheduleOnce(
+      'afterRender',
+      this,
+      this.checkTargetScrollOffset
+    );
   }),
 
   setupScrollOffset() {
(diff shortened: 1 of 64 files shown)


alexander-alvarez:
It's definitely not a bad thing, although I wouldn't have done some of the formatting they codified.
I'm was waiting to see what comes of https://github.com/emberjs/core-notes/blob/master/ember-cli/2017-06/june-29.md#prettier but as always you're one step ahead.

buschtoens:
Haha, yeah. I suffer from a real bad case of shiny new toy syndrome. :joy: 

It's true that there are a few places, where I would have styled differently, but the overall consistency outweighs these few for me.

Thank you for the link to the notes. Must have slipped my attention.

addon/components/lt-body.js
```diff
@@ -336,7 +335,12 @@ export default Component.extend({
       _scrollTo,
       scrollToRow,
       _scrollToRow
-    } = this.getProperties(['scrollTo', '_scrollTo', 'scrollToRow', '_scrollToRow']);
+    } = this.getProperties([
+      'scrollTo',
+      '_scrollTo',
+      'scrollToRow',
+      '_scrollToRow'
+    ]);
```
buschtoens: For example, I wouldn't write these destructuring `getProperties` statements anymore. Instead I'd go with multiple `get` lines.

I'll change that, when I'm back in the office.

addon/components/lt-column-resizer.js
```diff
@@ -21,7 +17,9 @@ export default Component.extend({
 
   $column: computed(function() {
     return $(this.get('element')).parent('th');
-  }).volatile().readOnly(),
+  })
+    .volatile()
+    .readOnly(),
```
buschtoens: That's also a bit weird. But that would go away by using `ember-decorators`.

addon/utils/css-styleify.js
```diff
@@ -1,17 +1,11 @@
 import Ember from 'ember';
 
-const {
-  isPresent,
-  String: {
-    dasherize,
-    htmlSafe
-  }
-} = Ember;
+const { isPresent, String: { dasherize, htmlSafe } } = Ember;
```
buschtoens: This also looks a bit weird. But ES6 module imports fix this as well.

Which reminds me - we should upgrade the code base to ES6 module imports. :grin:

fran-worley:
Given how out of date this now is, I'm going to close so we can focus on ember octane support.

Issue #12, opened 2016-04-20
Performance problems


SirZach:
Hi again,

I've noticed a performance problem and I haven't figured out what is causing it yet. I tried attaching the profile but github doesn't allow .cpuprofiles. You can easily reproduce it by profiling the demo table at http://offirgolan.github.io/ember-light-table/#/. The rendering speed isn't that noticeable since you're only rendering ~20 rows off the bat but even at ~150 rows, rendering speed is at ~2.5 seconds. I've reduced my table down to one row and have noticed in the profile that the one row creates about 3 large spikes that have a huge number of function calls. You can see spike after spike, for each row, if you profile the demo page.

I'm continuing to investigate. We really like `ember-light-table` but this performance issue is a show-stopper at the moment. Cheers,

Zach

offirgolan:
Hi Zach. In terms of performance, one of the biggest issues is not being able to render a non existent component. If you look at the code base, we are still forced to create and render each invisible component. This means that for each row, we actually render 3 (although 2 of which are just comments, we are still forced to create two additional views). I've spoken to @rwjblue about this and it seems to be a limitation with the component helper not being able to accept null as a component name. 

In terms of profiling, can you provide a snippet of your table setup so that way we are all profiling the same thing? I'm currently on vacation but I'll be able to look at this early next week.

SirZach:
```hbs
  {{#netuitive-table-new model=model columns=columns as |ntn|}}
    {{#light-table ntn.table tableActions=(hash checkAll=(action 'checkAll')) as |t|}}
      {{t.head
        onColumnClick=(action ntn.onColumnClick)
        iconAscending='fa fa-sort-asc'
        iconDescending='fa fa-sort-desc'
      }}

      {{t.body
        canSelect=false}}

    {{/light-table}}
  {{/netuitive-table-new}}
```

`netuitive-table-new` simply takes care of `init`ing the Table and making it available. For consistent profiling though, profiling the table on your demo site, http://offirgolan.github.io/ember-light-table/#/ shows exactly what I'm seeing locally. Thanks for replying even while on vacation!

SirZach:
Do you know off the top of your head where in the codebase it creates and renders each invisible component? I've poked through it quite a bit and I'm not sure I've seen that or know exactly what you mean.

offirgolan:
See addon/templates/components/lt-body.hbs

SirZach:
[...]

PR #483, opened 2017-10-02
[FEATURE occlusion]: Preliminary Vertical collection integration 

alexander-alvarez:
**WHY**
https://github.com/offirgolan/ember-light-table/issues/12 https://github.com/offirgolan/ember-light-table/issues/435
Especially https://github.com/offirgolan/ember-light-table/issues/12#issuecomment-318963414

**WHAT**
This PR introduces experimental integration with https://github.com/html-next/vertical-collection, as it is nearing a `1.0` in terms of stability. By setting the `occlusion=true` flag and setting a value for `estimatedRowHeight`, we deactivate `ember-scrollable` and use `vertical-collection` instead.

There is a new section in the cookbook `cookbook/occlusion-rendering` where an example is live

Some compromises with respect to functionality are made with this initial release so that it can get into people's hands and begin testing and playing with it. These are listed below:

**CAVEATS (To be turned into issues)**
- [ ] Must define widths for all columns -- Fixing this involves handling resize events, and making sure column widths stay synchronized across header / body tables
- [ ] Infinite render needs some TLC -- Fixing this involves usage/understanding of `scrollBuffer` property in ELT & vertical collection
- [ ] Fixed footer + occlusion rendering -- likely need to account for footer width when determining the width of the body as we did for the header.
- [ ] Native scrollbars (and 14px padding) are added -- integrating virtual scrollbars into `vertical collection` or `vertical collection` into `ember scrollable`
- [ ] Tests?

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "test": "ember try:each"
   },
   "dependencies": {
+    "@html-next/vertical-collection": "^1.0.0-beta.9",
     "ember-cli-babel": "^6.6.0",
     "ember-cli-htmlbars": "^2.0.2",
     "ember-cli-string-helpers": "^1.5.0",
@@ -34,7 +35,6 @@
     "ember-wormhole": "^0.5.1"
   },
   "devDependencies": {
-    "@html-next/vertical-collection": "^1.0.0-beta.8",
     "broccoli-asset-rev": "^2.4.5",
     "ember-ajax": "^3.0.0",
     "ember-cli": "2.13.2",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "ember-wormhole": "^0.5.1"
   },
   "devDependencies": {
+    "@html-next/vertical-collection": "^1.0.0-beta.8",
     "broccoli-asset-rev": "^2.4.5",
     "ember-ajax": "^3.0.0",
     "ember-cli": "2.13.2",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "ember-get-config": "^0.2.2",
     "ember-in-viewport": "2.1.1",
     "ember-scrollable": "^0.4.7",
-    "ember-truth-helpers": "1.3.0",
+    "ember-truth-helpers": "^2.0.0",
     "ember-wormhole": "^0.5.1"
   },
   "devDependencies": {
@@ -64,7 +64,7 @@
     "ember-font-awesome": "^3.0.5",
     "ember-load-initializers": "^1.0.0",
     "ember-native-dom-helpers": "^0.5.0",
-    "ember-one-way-controls": "2.0.0",
+    "ember-one-way-controls": "^2.0.0",
     "ember-owner-test-utils": "^0.1.2",
     "ember-resolver": "^4.0.0",
     "ember-responsive": "^2.0.4",
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ A lightweight contextual component based table addon that follows Ember's action
 - Easy table manipulation
 - Easy override to table header, body, and footer
 - Contextual component for header, body, and footer, as well as loading, no data, and expanded row
+- **EXPERIMENTAL** Occlusion rendering leveraging [vertical-collection](https://github.com/html-next/vertical-collection). See [Demo](http://offirgolan.github.io/ember-light-table/#/cookbook/occlusion-rendering).
 
 ## Installation
 
--- a/addon/components/light-table.js
+++ b/addon/components/light-table.js
@@ -38,7 +38,7 @@ function intersections(array1, array2) {
 
 const LightTable = Component.extend({
   layout,
-  classNameBindings: [':ember-light-table'],
+  classNameBindings: [':ember-light-table', 'occlusion'],
   attributeBindings: ['style'],
 
   media: service(),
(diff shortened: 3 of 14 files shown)


offirgolan:
This is an amazing start @alexander-alvarez. Thanks for putting in the time to make this happen! Im super excited for the next iterations 😄

mkay581:
> Must define widths for all columns -- Fixing this involves handling resize events, and making sure column widths stay synchronized across header / body tables

Will this fix what I attempted to in #298?

alexander-alvarez:
@mkay581 quite possibly, as the implementation will likely have to inspect the DOM like you do in your PR

alexander-alvarez:
woopss.. fat finger..
@buschtoens  did you want to take a look?

KeithClinard:
I am having two issues with this branch.

The table header is respecting the column widths that were specified, but the table cells are not.
This causes the columns to be misaligned.

The other issue is that the header appears to be a few pixels taller than it has room for, leading to a scroll bar in the the top right for the header.
This effect becomes more pronounced when more padding is added to the header cells.

Both of these issues can be reproduced easily by using the `occluded-table` in the dummy app.

Here is a screenshot of what I am seeing. (Borders added for emphasis)
image

alexander-alvarez:
thanks @KeithClinard will take a look

KeithClinard:
I just managed to get this to render properly.
It seems that the `vertical-collection` and `occluded-content` html nodes broke the table rendering.

I was able to get it to work by tweaking the `display` property of these nodes to use table styles.
```css
vertical-collection {
  display: table;
  table-layout: fixed;
}
vertical-collection occluded-content:first-of-type {
  display: table-caption;
}
```

There is likely a cleaner way to write the css, but I thought I would give you what I have.

alexander-alvarez:
@KeithClinard  thanks for the investigation and the fix, I've incorporated your changes. Check them out.

As to:

>The other issue is that the header appears to be a few pixels taller than it has room for, leading to a scroll bar in the the top right for the header.
This effect becomes more pronounced when more padding is added to the header cells.

I think you have to style the height of the header row appropriately so it doesn't scroll (padding & border widths included), but I could be wrong and there could be a way to have this not be a requirement.

buschtoens:
I _finally_ got around to try this out. First of all: absolutely amazing work! Thank you so much. :tada: :heart: 

I dropped this into a rather complex table that has to display up to 500 rows simultaneously. Previously we incrementally pushed rows into the table to not freeze the main UI thread. I can confirm that I can now just throw in all 500 rows at once and it _just works_.

I haven't yet discovered any styling issues. Functionality-wise I only discovered that `scrollTo` and `scrollToRow` do not work which makes sense since the rely on ember-scrollable.

I'll give this further test drives in the coming days and check back in. But so far, I can"t :+1: this enough!

buschtoens:
I found a minor issue, but I suspect that it's an issue in `vertical-collection` itself: when the table is created while the browser tab is hidden, this error is thrown, but does no apparent harm:

```
Uncaught Error: Attempted to remove a handler from an unknown element or an element with no handlers
    at ScrollHandler.removeScrollHandler (-private.js:1756)
    at removeScrollHandler (-private.js:1817)
    at Class.willDestroy [as _super] (component.js:199)
    at Class.willDestroy (debug-mixin.js:82)
    at Class.superWrapper [as willDestroy] (ember-utils.js:426)
    at invoke (backburner.js:274)
    at Queue.flush (backburner.js:153)
    at DeferredActionQueues.flush (backburner.js:345)
    at Backburner.end (backburner.js:455)
    at Backburner.run (backburner.js:539)
```

```js
// assets/addon-tree-output/modules/vertical-collection/-private.js
  ScrollHandler.prototype.removeScrollHandler = function removeScrollHandler(element, handler) {
    var index = this.elements.indexOf(element);
    var elementCache = this.handlers[index];
    // TODO add explicit test
    if (elementCache && elementCache.handlers) {
      var _index = elementCache.handlers.indexOf(handler);

      if (_index === -1) {
        throw new Error('Attempted to remove an unknown handler');
      }

      elementCache.handlers.splice(_index, 1);

      // cleanup element entirely if needed
      // TODO add explicit test
      if (!elementCache.handlers.length) {
        _index = this.elements.indexOf(element);
        this.handlers.splice(_index, 1);
        this.elements.splice(_index, 1);

        this.length--;
        this.maxLength--;

        if (this.length === 0) {
          this.isPolling = false;
        }

        // TODO add explicit test
        if (this.isUsingPassive) {
          element.removeEventListener('scroll', elementCache.passiveHandler, { capture: true, passive: true });
        }
      }
    } else {
      throw new Error('Attempted to remove a handler from an unknown element or an element with no handlers');
    }
  };
```

package.json
```diff
@@ -30,10 +30,11 @@
     "ember-get-config": "^0.2.2",
     "ember-in-viewport": "2.1.1",
     "ember-scrollable": "^0.4.7",
-    "ember-truth-helpers": "1.3.0",
+    "ember-truth-helpers": "^2.0.0",
     "ember-wormhole": "^0.5.1"
   },
   "devDependencies": {
+    "@html-next/vertical-collection": "^1.0.0-beta.8",
```
buschtoens: Is this intentionally only a `devDependency`, because not every user will want to try occlusion rendering just yet?

alexander-alvarez:
@buschtoens  sounds related to https://github.com/html-next/vertical-collection/pull/140

package.json
alexander-alvarez: I think this was a blunder on my part... will update when I update to `beta.9`

alexander-alvarez:
Status update... generally well received & thanks to help from @KeithClinard we've gotten to a pretty good place...

I'm waiting for a `1.0.0.beta.9` release of `vertical-collection` because it has breaking changes, then we should be able to roll this out and get people to experiment with it.

alexander-alvarez:
https://github.com/html-next/vertical-collection/releases/tag/v1.0.0-beta.9 was just released, so I updated this branch to reflect the API changes.

we may explore moving the vertical collection into `lt-scrollable` as we're figuring out the optimal mechanics (should we use it as a tag-less component?) and we're connecting full set of APIs offered by vertical collection.

Please take a look

buschtoens:
I'm super happy with this PR and would love, if we'd merge it. Since it's behind a flag, we can change some finer details later. :smile: 

If you're okay with it, I'd merge and release, so I can properly test drive it in my apps. :grin:

alexander-alvarez:
The green button is :point_down:

efx:
thanks for all the work @alexander-alvarez! (and reviewers like @buschtoens ). :) can someone merge this please? I would like to test this in our application that has numerous large tables.

buschtoens:
Release as `v1.12.0` :tada:

Issue #478, opened 2017-08-21
Customizable sort icon component


RedTn:
It would be nice if there were some properties we can pass into the content of this element here:
[https://github.com/offirgolan/ember-light-table/blob/master/addon/templates/components/columns/base.hbs#L6](https://github.com/offirgolan/ember-light-table/blob/master/addon/templates/components/columns/base.hbs#L6)
Materialize expects the element to be defined like this:
`<i class="material-icons">add</i>`
[http://materializecss.com/icons.html](http://materializecss.com/icons.html)

RedTn:
- [ ]

alexander-alvarez:
Should be able to accomplish this leveraging
http://offirgolan.github.io/ember-light-table/docs/classes/t.head.html#property_iconAscending
http://offirgolan.github.io/ember-light-table/docs/classes/t.head.html#property_iconDescending
and
http://offirgolan.github.io/ember-light-table/docs/classes/t.head.html#property_iconSortable

see the examples in the example app:
![image](https://user-images.githubusercontent.com/11779993/29539300-4427b23a-8697-11e7-875d-20298287a5a6.png)

RedTn:
Right, but materialize also requires the content in that **i** tag in order to determine the icon to use.
So in my example, it requires:
`<i ...>add</i>`.
Ideally, we could do something like:
`{{t.head
...
iconSortable="material-icons"
iconSortableValue="add"}}`

alexander-alvarez:
got it, now I understand.. I've never heard of anything like that, but I understand how it works.

In the meantime you can define your own column component that gets called [here](https://github.com/offirgolan/ember-light-table/blob/master/addon/templates/components/columns/base.hbs#L2).
You can generate it with `ember g column-type <your_column_type>`
and set the http://offirgolan.github.io/ember-light-table/docs/classes/Column.html#property_type property on your columns passed into to light table.

I'm not sure if we would want to yet another property for icons or maybe make the icon component customizable all together. 

@buschtoens  any thoughts?

buschtoens:
> I've never heard of anything like that, but I understand how it works.

Count me in. :smile: 

> I'm not sure if we would want to yet another property for icons or maybe make the icon component customizable all together.

I'm on the same page. We should rather add an `iconComponet` to the `TableHeaderMixin` and render it instead of the `<i>` tag, that was already getting a bit too complex anyways.

Absolutely in favor! :+1:

RedTn:
[...]

PR #489, opened 2017-10-09
Feat/custom icon components

RedTn:
This a a PR for an issue I submitted for customizable sort components: #478

### DIFF
--- a/addon/mixins/table-header.js
+++ b/addon/mixins/table-header.js
@@ -107,10 +107,8 @@ export default Mixin.create({
   iconDescending: '',
 
   /**
-   * See `iconSortable, iconAsending, or iconDescending.  Custom sorting component
-   * name to use instead of the default `<i class="lt-sort-icon"></i>` template. Uses
-   * the same properties.
-   * 
+   * Custom sorting component name to use instead of the default `<i class="lt-sort-icon"></i>` template.
+   * See `iconSortable, iconAsending, or iconDescending.
    * @property iconComponent
    * @type {String}
    * @default false
--- a/addon/mixins/table-header.js
+++ b/addon/mixins/table-header.js
@@ -107,6 +107,17 @@ export default Mixin.create({
   iconDescending: '',
 
   /**
+   * See `iconSortable, iconAsending, or iconDescending.  Custom sorting component
+   * name to use instead of the default `<i class="lt-sort-icon"></i>` template. Uses
+   * the same properties.
+   * 
+   * @property iconComponent
+   * @type {String}
+   * @default false
+   */
+  iconComponent: null,
+
+  /**
    * ID of main table component. Used to generate divs for ember-wormhole
    * @type {String}
    */
@@ -117,8 +128,8 @@ export default Mixin.create({
   subColumns: computed.readOnly('table.visibleSubColumns'),
   columns: computed.readOnly('table.visibleColumns'),
 
-  sortIcons: computed('iconSortable', 'iconAscending', 'iconDescending', function() {
-    return this.getProperties(['iconSortable', 'iconAscending', 'iconDescending']);
+  sortIcons: computed('iconSortable', 'iconAscending', 'iconDescending', 'iconComponent', function() {
+    return this.getProperties(['iconSortable', 'iconAscending', 'iconDescending', 'iconComponent']);
   }).readOnly(),
 
   init() {
--- a/addon/mixins/table-header.js
+++ b/addon/mixins/table-header.js
@@ -108,7 +108,7 @@ export default Mixin.create({
 
   /**
    * Custom sorting component name to use instead of the default `<i class="lt-sort-icon"></i>` template.
-   * See `iconSortable, iconAsending, or iconDescending.
+   * See `iconSortable`, `iconAsending`, or `iconDescending`.
    * @property iconComponent
    * @type {String}
    * @default false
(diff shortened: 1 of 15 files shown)


alexander-alvarez:
@RedTn sorry for the delay in communication. I had a weird week with holidays, vacation, and other stuff...
Will try to get to it early this week, but on first glance looks good.

RedTn:
Feel free to let me know if anything needs to be changed/updated

RedTn:
@alexander-alvarez Checking up on this PR, thanks alexander!

addon/templates/components/columns/base.hbs
```diff
@@ -2,7 +2,9 @@
   {{component column.component column=column table=table tableActions=tableActions extra=extra sortIcons=sortIcons}}
 {{else}}
   {{label}}
-  {{#if sortIconProperty}}
+  {{#if (and sortIcons.iconComponent sortIconProperty)}}
+    {{component sortIcons.iconComponent sortIcons=sortIcons sortIconProperty=sortIconProperty}}
```
alexander-alvarez: would you mind adding a test to verify the arguments (`sortIcons=sortIcons sortIconProperty=sortIconProperty`) get passed to the component declared by  `iconComponent` when we render a table like you do in the demo app:

```diff
+  {{t.head
+    onColumnClick=(action 'onColumnClick')
+    iconSortable='unfold_more'
+    iconAscending='keyboard_arrow_up'
+    iconDescending='keyboard_arrow_down'
+    iconComponent='materialize-icon'
+    fixed=true
+  }}
```

You can steal some ideas from
https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/customization-with-components-test.js#L257..L281

addon/mixins/table-header.js
```diff
@@ -107,6 +107,17 @@ export default Mixin.create({
   iconDescending: '',
 
   /**
+   * See `iconSortable, iconAsending, or iconDescending.  Custom sorting component
```
alexander-alvarez: I would change the order
```
Custom sorting component name to use instead of the default `<i class="lt-sort-icon"></i>` template. 
See `iconSortable`, `iconAsending`, or `iconDescending` for more info.
```
> Uses the same properties.

It's unclear what you mean by this sure what you mean by this

tests/integration/components/lt-head-test.js
```diff
@@ -92,6 +94,34 @@ test('render sort icons', function(assert) {
   assert.notOk(hasClass(sortIcon, 'fa-sort-asc'));
 });
 
+test('custom iconComponent has arguments', function(assert) {
+  const iconSortable = 'unfold_more';
```
alexander-alvarez: `assert.expect(6)`
otherwise the test will pass if the component doesn't get rendered

RedTn: @alexander-alvarez Will add this soon, thanks alexander I'm learning quite a bit from this PR

addon/mixins/table-header.js
```diff
@@ -107,6 +107,15 @@ export default Mixin.create({
   iconDescending: '',
 
   /**
+   * Custom sorting component name to use instead of the default `<i class="lt-sort-icon"></i>` template.
+   * See `iconSortable, iconAsending, or iconDescending.
```
buschtoens: Minor formatting error with the backticks here.

```js
See `iconSortable`, `iconAsending`, or `iconDescending`.
```

addon/templates/components/columns/base.hbs
```diff
@@ -2,7 +2,9 @@
   {{component column.component column=column table=table tableActions=tableActions extra=extra sortIcons=sortIcons}}
 {{else}}
   {{label}}
-  {{#if sortIconProperty}}
+  {{#if (and sortIcons.iconComponent sortIconProperty)}}
```
buschtoens: Can't remember, if we talked about that. but this way the `iconComponent` would only get rendered when there is a [`sortIconProperty`](https://github.com/RedTn/ember-light-table/blob/7975553815eda87054da4f0ec85d248c9cdedc77/addon/components/columns/base.js#L82-L92), i.e. when either of `sortable` of `sorted` is `true`.

I don't have strong feelings either way, since I can't come up with a scenario where you would want to render an `iconComponent` when you don't have any icons to display. Just wanted to have mentioned it. :smile:

RedTn: I remember I ran into this issue in the demo where the avatar column was displaying the sort component when sortable: 'false'.
Because the iconComponent is being specified in the hbs file on the table, all the columns will utilize the iconComponent, so we need a check if the user doesn't want a column to be sortable.

buschtoens: Yup, but some users might argue that they'd want their `iconComponent` to display a "not sortable" icon and that it's up to the user to handle the "not sortable" state instead of ELT doing that for the user.

For instance:

```hbs
{{#if sortIconProperty}}
  <i class="lt-sort-icon material-icons" style="height:0px;">{{get sortIcons sortIconProperty}}</i>
{{/if}}
```

RedTn: Understood, I see the use case. Because you have no strong feelings I guess I'll leave the base.hbs as is for now.

buschtoens:
Thanks so much @RedTn :heart:

Issue #494, opened 2017-10-13
`droppable` attribute is not respected.


akshay-kr:
Docs mentions that `droppable` attribute defines `Whether the column is a valid drop target.`.
I tried setting it to `false` for a column but still it gets displaced by dropping another column on it.

I think we should have a check for `droppable` attribute after below line,
https://github.com/offirgolan/ember-light-table/blob/master/addon/mixins/draggable-column.js#L116
```javascript
if(targetColumn.droppable){ 
  //then proceed
}
```
If we agree I can create a PR for this.

PR #496, opened 2017-10-13
Fix-Add droppable attribute check on target column before dropping source column

akshay-kr:
`droppable` attribute is not respected when  dropping source column over target column where the attribute is set to false.

Just added a check on this PR for `droppable` attribute on target column. 

Related Issue : #494


alexander-alvarez:
Thanks for this! Seems like a quite a big woopsie :face_with_head_bandage: 
Will update here when the version including this change is out

Issue #508, opened 2017-11-08
Vertical Collection: Wire up `firstReach`, `lastReached` events to light table's own


alexander-alvarez:
See https://github.com/html-next/vertical-collection#actions for reference

And our analgous http://offirgolan.github.io/ember-light-table/docs/classes/t.body.html#event_onScrolledToBottom.

Similar thing goes for `onScroll` assuming that functionality is useful

buschtoens:
Closed in #514. :smile:

PR #514, opened 2017-11-23
fix(occlusion): pass bufferSize based on scrollBuffer

### LINKED ISSUE #485: Vertical-Collection: Fix Infinite Render / Scrollbuffer
`scrollBuffer` in vertical collection is item count, and in ember light table is in pixels.

Luckily we allow the definition of `estimatedRowHeight`, so we can use that to translate from pixels to item count

buschtoens:
Closes #485.

### DIFF
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -195,6 +195,17 @@ export default Component.extend({
   scrollBuffer: 500,
 
   /**
+   * @property scrollBufferRows
+   * @type {Number}
+   * @default 500 / estimatedRowHeight
+   */
+  scrollBufferRows: computed('scrollBuffer', 'sharedOptions.estimatedRowHeight', function() {
+    return Math.ceil(
+      this.get('scrollBuffer') / (this.get('sharedOptions.estimatedRowHeight') || 1)
+    );
+  }),
+
+  /**
    * @property useVirtualScrollbar
    * @type {Boolean}
    * @default false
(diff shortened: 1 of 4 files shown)


buschtoens:
Looking at the [actions made available by `vertical-collection`](https://github.com/html-next/vertical-collection#actions), this PR (in it's current form) doesn't really gain us anything yet. While the `bufferSize` is correctly calculated, the `onScrolledToBottom` action is not called yet.

Therefore I'll keep this WIP and also try to solve #508 in this PR.

buschtoens:
Since we want to trigger `onScrolledToBottom` with `scrollBuffer` in mind, we should probably use `lastVisibleChanged` instead of `lastReached`.

`onScroll` could be shimmed with `firstVisibleChanged`. However the `onScroll` action wouldn't be called for *every* scroll event, but I think that's a worthwhile trade-off.`

buschtoens:
Okay, so I got the actions wired up, but we have a problem.

`vertical-collection` is not happy about the `{{#if isLoading}}` construct... Essentially Glimmer needs a static element to render the contents of `{{#if isLoading}}` into. If you wrap it with a plain `<div>`, everything works fine, but of course that bricks the layout...

There is a private [`{{#-in-element someElement}}` helper](https://github.com/glimmerjs/glimmer-vm/pull/331) that takes the block and renders it into `someElement`. This would fix our problem, *however* we need something like `{{#-after-element someElement}}`, which doesn't exist yet.

buschtoens:
While I'm trying to fix this (or get people to fix it :stuck_out_tongue_winking_eye:), we can still ship out a new version, since occlusion has to be enabled with a flag and tables with occlusion enabled and infinite scroll were essentially broken anyway. :smile: 

I'll drop a line in the release notes that instruct users to not use any conditionals in the the yielded block.

Issue #450, opened 2017-07-05
Externally provided models


buschtoens:
I had a quick chat with @shuriu in our #e-light-table Slack channel.

He asked what the preferred pattern for externally provided models is:

> hey, is there an accepted way to load the records from the route instead of the table component?
>I just realised that instatiating a table via `new Table(get(this, 'columns'), get(this, 'model'))`
> will not work :disappointed:

Depending on the scenario, there are two options I would suggest:

#### Computed Property

```js
model: null,

table: computed('model', function() {
  return new Table(get(this, 'columns'), get(this, 'model'), { enableSync: true });
})
```

The obvious problem with this is, that whenever the `model` array changes (the reference / identity, not mutations), the `table` is re-created, thus losing all state.

#### `didReceiveAttrs` hook

```js
model: null,

table: null,

init() {
  this._super(...arguments);

  set(this, 'table', new Table(get(this, 'columns'), get(this, 'model') || []));
},

didReceiveAttrs: diffAttrs('model', function(changedAttrs, ...args) {
  this._super(...args);

  if (changedAttrs && changedAttrs.model) {
    get(this, 'table').setRows(changedAttrs.model[1] || []);
  }
}),
```

We're using the excellent [`ember-diff-attrs`](https://github.com/workmanw/ember-diff-attrs) here. The advantage with this pattern is, that the very same `table` is re-used throughout the lifespan of the component.

One drawback is that `enableSync` cannot be used. Once #414 is fixed, this is no problem any more.

RedTn:
@buschtoens does this just need a cookbook example or is there some sort of enhancement needed? I can write the cookbook since I'm using this pattern for my ELTs.

buschtoens:
Thanks for pinging me on Slack!  Yes, this is purely cookbook. 

Things that might be of interest and can be explored:

- changing the externally provided model, e.g. updating the attr
- augmenting the externally provided model via infinite scroll

The scenario here is that the initial rows are loaded via the route's `model` hook and then additional rows are added on demand.

timmyomahony:
Just to chime in for anyone else this might help. I came to ELT with the traditional route-based model being passed down to the component so I was a bit confused when setting up my tables  using the existing cookbook. I've got a model that's getting filtered via a queryparam. 

Here's my mixin based off @buschtoens code above, using `ember-did-change-attrs` instead:

```
import Mixin from '@ember/object/mixin';
import Table from 'ember-light-table';
[...]

PR #527, opened 2017-12-17
docs(cookbook): add sample for externally provided model

RedTn:
Closes https://github.com/offirgolan/ember-light-table/issues/450


tests/dummy/app/components/cookbook/external-model-table.js
```diff
@@ -0,0 +1,97 @@
...
+  }),
+
+  canLoadMore: true,
+  enableSync: true,
+
+  meta: null,
+  table: null,
+  rows: A([]),
+
+  didReceiveAttrs: diffAttrs('model', function(changedAttrs, ...args) {
```
RedTn: Might be worth switching to [ember-did-change-attrs](https://github.com/workmanw/ember-did-change-attrs) at some point but didn't now since it's still a WIP.

cristinawithout:
I don't think you want this example to set `rows: A([])` on the table component.

It's likely people will want to reuse the component they create with an external data source in more than one place across their application. Setting it this way will make `rows` shared across all instances of the table. (So when rows change in one table, it will change in every table.) Should probably set the value somewhere like init().

RedTn:
@buschtoens 
Is this still good to be merged once I fix the merge conflicts? The comment from @cristinawithout has already been addressed in the commits I can't resolve the comment on the ui :(

PR #535, opened 2018-01-11
[WIP] fix(Column): `minResizeWidth`

### LINKED ISSUE #534: Make `minResizeWidth` take a string like `width` or vice-versa
Currently the `Column` class requires `minResizeWidth` to be a number (in px)  while `width` is provided as a string like so `width: '100px'. I think it would be good to have these two properties work the same for consistency sake. For backward compatibility, we could allow a number or a string for both.

cojomojo:
Closes #534

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -1,7 +1,8 @@
 import { A as emberArray, makeArray } from '@ember/array';
+import { warn } from '@ember/debug';
 import EmberObject, { computed } from '@ember/object';
-import { isEmpty } from '@ember/utils';
 import { guidFor } from '@ember/object/internals';
+import { isEmpty, typeOf } from '@ember/utils';
 import fixProto from 'ember-light-table/utils/fix-proto';
 
 /**
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
 
   /**
    * The minimum width (in px) that this column can be resized to.
-   * @property minResizeWidth
+   * @property _minResizeWidth
    * @type {Number}
    * @default 0
+   * @private
    */
-  minResizeWidth: 0,
+  _minResizeWidth: 0,
+
+  /**
+   * The minimum width that this column can be resized to.
+   * @property minResizeWidth
+   * @type {String}
+   * @default '0px'
+   */
+  minResizeWidth: computed({
+    get() {
+      return this.get('_minResizeWidth');
+    },
+    set(key, value) {
+      if (typeOf(value) === 'string') {
+        let [, quantity, units] = value.match(/(\d+)(px)?/) || [];
+        if (units !== 'px') {
+          warn('`value` attribute is interpreted as px regardless of provided units', {
+            id: 'ember-light-table.classes.Column'
+          });
+        }
+        this.set('_minResizeWidth', Number(quantity));
+      } else {
+        this.set('_minResizeWidth', value);
+      }
+    }
+  }),
 
   /**
    * The parent column (or group) for this sub-column.
--- a/tests/unit/classes/column-test.js
+++ b/tests/unit/classes/column-test.js
@@ -78,6 +78,31 @@ test('CP - visibleSubColumns', function(assert) {
   assert.equal(col.get('visibleSubColumns.length'), 0);
 });
 
+test('CP - minResizeWidth', function(assert) {
+  let col = new Column({
+    minResizeWidth: '100px'
+  });
+  assert.ok(col);
+  assert.equal(col.get('_minResizeWidth'), 100);
+  col = new Column({
+    minResizeWidth: '100%'
+  });
+  assert.ok(col);
+  assert.equal(col.get('_minResizeWidth'), 100);
+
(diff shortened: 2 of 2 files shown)


addon/classes/Column.js
```diff
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
...
+    },
+    set(key, value) {
+      if (typeOf(value) === 'string') {
+        let [, quantity, units] = value.match(/(\d+)(px)?/) || [];
+        if (units !== 'px') {
+          warn('`value` attribute is interpreted as px regardless of provided units', {
+            id: 'ember-light-table.classes.Column'
+          });
+        }
+        this.set('_minResizeWidth', Number(quantity));
```
buschtoens:
```js
return this.set('_minResizeWidth', Number(quantity));
```

> I wrote a unit test for the work, but I am unsatisfied with it because I had to do `col.get('_minResizeWidth')` and not `col.get('minResizeWidth')` because the latter was returning undefined. Any idea why that could be?

This is why. When using the `computed({ get(k) { ... }, set(k , v) { ... } })` form, the setter method needs to return a value as well, as per the [API docs](https://emberjs.com/api/ember/2.18/classes/@ember%2Fobject/methods/computed?anchor=computed): 

> The `set` function should accept two parameters, `key` and `value`. The value returned from `set` will be the new value of the property.

By prepending `return` to `this.set(k, v)` you return the value passed to it, since [`this.set(k, v)` returns `v`](https://emberjs.com/api/ember/2.18/classes/@ember%2Fobject/methods/set?anchor=set).

addon/classes/Column.js
```diff
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
...
+      } else {
+        this.set('_minResizeWidth', value);
```
buschtoens:
```js
return this.set('_minResizeWidth', value);
```

Also, what about other non-string values? What if I pass `NaN`, `{}`, `false`, `null`, ...? I think we should handle these cases a bit more explicitly and maybe even throw assertions / `TypeError`s for some scenarios.

addon/classes/Column.js
```diff
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
...
+  _minResizeWidth: 0,
+
+  /**
+   * The minimum width that this column can be resized to.
+   * @property minResizeWidth
+   * @type {String}
+   * @default '0px'
    */
-  minResizeWidth: 0,
+  minResizeWidth: computed({
```
buschtoens: Missing dependent key `_minResizeWidth`.

In practice this won't cause any problems yet, since `_minResizeWidth` is private API and exclusively managed by the `minResizeWidth` computed property. However, that could possibly change in the future and the developer making that change *then* might not remember to add that key.

addon/classes/Column.js
```diff
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
...
+    get() {
+      return this.get('_minResizeWidth');
+    },
+    set(key, value) {
+      if (typeOf(value) === 'string') {
```
buschtoens:
```js
typeof value === 'string'
```

is a bit faster. The only advantage of [`typeOf(item)`](https://github.com/emberjs/ember.js/blob/v2.18.0/packages/ember-runtime/lib/utils.js#L62) is that it correctly detects `String` object instances as `string`, i.e. `typeOf(new String('foo')) === 'string'`.

Personally idc a lot about this scenario, but that just might be because this has never bitten me yet. Keep this line whatever you feel the most comfortable with. :wink:

addon/classes/Column.js
```diff
@@ -129,11 +130,37 @@ export default class Column extends EmberObject.extend({
...
+        let [, quantity, units] = value.match(/(\d+)(px)?/) || [];
+        if (units !== 'px') {
+          warn('`value` attribute is interpreted as px regardless of provided units', {
+            id: 'ember-light-table.classes.Column'
+          });
```
buschtoens:
```js
const [, quantity, units] = value.match(/^\s*(\d+)([a-z]+|%)?\s*$/) || [];
assert(
  `'minResizeWidth' can only interpret px unit values. You passed '${quantity}${unit}'.`,
  isEmpty(units) || units === 'px'
);
```

Being more restrictive here can help prevent subtle bugs. We should throw an assertion here, since `1rem` would be interpreted as `1px` in the original code, which does not make sense.

buschtoens:
It really nags me, that we can't get other units to work... :slightly_frowning_face: 

But at least supporting `px` as a string is better that crashing for no good reason. :+1:

addon/classes/Column.js
cojomojo: @buschtoens The only concern I have about throwing assertions for `TypeError`s is that it could cause people who are incorrectly using `minResizeWidth` currently to receive an exception which may negatively affect their application. Now you can argue that since they are already using it incorrectly, this is actually better, but it is something to think about. 

I plan on making the suggested changes this upcoming week.

PR #539, opened 2018-02-02
Upgrade to 2.18

snewcomer:
Upgrade to 2.18

Definitely might break something :)

PR #531, opened 2017-12-19
fix tests

buschtoens:
Updates to ember-cli 2.17 and fixes and re-enables all tests.

PR #546, opened 2018-02-22
Ember 3.0 and working tests

buschtoens:
Closes #539. Closes #531.


buschtoens:
The infinity scroll stuff seems to be broken now. It only fetches new data, when the infinity element is in viewport. Scrolling apparenty does not trigger anything.

PR #548, opened 2018-02-24
Update in viewport to use IntersectionObserver

### LINKED ISSUE #555: Can you please resolve the deprecation ember-cli-version-checker ?
The message keeps showing when moving the ember light table into an engine.
`ember-in-viewport` is the dependency of `ember-light-table`, isn't it?

`DEPRECATION: An addon is trying to access project.nodeModulesPath. This is not a reliable way to discover npm modules. Instead, consider doing: require("resolve").sync(something, { basedir: project.root }). Accessed from:   DependencyVersionChecker.NPMDependencyVersionChecker (/Users/bzhang2/Project/mg-ui/node_modules/ember-in-viewport/node_modules/ember-cli-version-checker/src/npm-dependency-version-checker.js:11:32)`

snewcomer:
Trying to get IntersecitonObserver working.

- One option is to make the loader the height of the loading icon.  This would avoid jumps b/w reaching `lt-infinity` and showing the loading spinner
- noticing that bottom/top tolerance for IntersectionObserver is not being respected.

close #555

### DIFF
--- a/addon/components/lt-infinity.js
+++ b/addon/components/lt-infinity.js
@@ -1,6 +1,4 @@
 import Component from '@ember/component';
-import { observer } from '@ember/object';
-import { run } from '@ember/runloop';
 import layout from '../templates/components/lt-infinity';
 import InViewportMixin from 'ember-in-viewport';
 
@@ -26,7 +24,7 @@ export default Component.extend(InViewportMixin, {
         left: width,
         right: width,
         bottom: scrollBuffer,
-        top: scrollBuffer
+        top: 0
       },
       scrollableArea: scrollableContent
     });
@@ -34,40 +32,9 @@ export default Component.extend(InViewportMixin, {
 
   willDestroyElement() {
     this._super(...arguments);
-    this._cancelTimers();
   },
 
   didEnterViewport() {
-    this._debounceScrolledToBottom();
-  },
-
-  didExitViewport() {
-    this._cancelTimers();
-  },
-
-  scheduleScrolledToBottom: observer('rows.[]', 'viewportEntered', function() {
-    if (this.get('viewportEntered')) {
-      /*
-       Continue scheduling onScrolledToBottom until no longer in viewport
-       */
-      this._scheduleScrolledToBottom();
-    }
-  }),
-
-  _scheduleScrolledToBottom() {
-    this._schedulerTimer = run.scheduleOnce('afterRender', this, this._debounceScrolledToBottom);
-  },
-
-  _debounceScrolledToBottom(delay = 100) {
-    /*
-     This debounce is needed when there is not enough delay between onScrolledToBottom calls.
-     Without this debounce, all rows will be rendered causing immense performance problems
-     */
-    this._debounceTimer = run.debounce(this, this.sendAction, 'onScrolledToBottom', delay);
-  },
-
-  _cancelTimers() {
-    run.cancel(this._schedulerTimer);
-    run.cancel(this._debounceTimer);
+    this.sendAction('onScrolledToBottom');
   }
 });
--- a/addon/components/lt-infinity.js
+++ b/addon/components/lt-infinity.js
@@ -20,7 +20,6 @@ export default Component.extend(InViewportMixin, {
 
     this.setProperties({
       viewportSpy: true,
-      viewportUseIntersectionObserver: false,
       viewportTolerance: {
         left: width,
         right: width,
--- a/addon/components/lt-infinity.js
+++ b/addon/components/lt-infinity.js
@@ -1,6 +1,4 @@
 import Component from '@ember/component';
-import { observer } from '@ember/object';
-import { run } from '@ember/runloop';
 import layout from '../templates/components/lt-infinity';
 import InViewportMixin from 'ember-in-viewport';
 
@@ -11,7 +9,7 @@ export default Component.extend(InViewportMixin, {
 
   rows: null,
   scrollableContent: null,
-  scrollBuffer: 0,
+  scrollBuffer: 50,
 
   didInsertElement() {
     this._super(...arguments);
@@ -32,38 +30,7 @@ export default Component.extend(InViewportMixin, {
     });
   },
 
-  willDestroyElement() {
-    this._super(...arguments);
-    this._cancelTimers();
-  },
-
   didEnterViewport() {
-    this._scrolledToBottom();
-  },
-
-  didExitViewport() {
-    this._cancelTimers();
-  },
-
-  scheduleScrolledToBottom: observer('rows.[]', 'viewportEntered', function() {
-    if (this.get('viewportEntered')) {
-      /*
-       Continue scheduling onScrolledToBottom until no longer in viewport
-       */
-      this._scheduleScrolledToBottom();
-    }
-  }),
-
-  _scheduleScrolledToBottom() {
-    this._schedulerTimer = run.scheduleOnce('afterRender', this, this._scrolledToBottom);
-  },
-
-  _scrolledToBottom() {
     this.sendAction('onScrolledToBottom');
-  },
-
-  _cancelTimers() {
-    run.cancel(this._schedulerTimer);
-    run.cancel(this._Timer);
   }
 });
--- a/addon/components/lt-infinity.js
+++ b/addon/components/lt-infinity.js
@@ -38,7 +38,7 @@ export default Component.extend(InViewportMixin, {
   },
 
   didEnterViewport() {
-    this.sendAction('onScrolledToBottom');
+    this._scrolledToBottom();
   },
 
   didExitViewport() {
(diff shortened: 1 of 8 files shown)


buschtoens:
I suspect that I bricked this stuff with 8c997038979416234685045ba15f91eb38230012 and felt like d0bcf7f226a21361b368011fb3937bf9afdaf8fc made it less severe.

I also tried this, but with `position: absolute` in order to not add anything to the layout.

https://github.com/offirgolan/ember-light-table/pull/548/commits/07deac2bdbc0bb890443127eadea8ece4763a56d#diff-696281ec0167c1756fc2a44bfc72aba9R67

While making it better, it was still flaky.

buschtoens:
Ideally, I would make occlusion rendering the default behavior and remove the non-occlusion mode. Then we could rely exclusively on the vertical-collection actions.

Obviously this will require a major release.

snewcomer:
@buschtoens Thanks for the context!  I see your point about not adding more to the layout. What would it take to have occlusion rending the default?

Also just a little more context on this PR.  First thing I noticed was that the scrolling behavior wasn't working correctly.  Perhaps it is working fine in other contexts.  In the demo, the infinity scrolling is now working in Firefox and Chrome with this PR b/c it is using the IntersectionObserver.

Also the `virtualScrollBar` was doing something funky (calling didInsertElement) twice, and messing with the in-viewport addon.

Lastly, this PR is not working in Safari (using rAF) b/c the scrollBuffer is 500 (> than the height of the scrolling area).  If the scrollBuffer was like 100, this would work across browsers.

snewcomer:
@buschtoens Actually I think the big thing I am noticing is that when using IntersectionObserver, there is some timing related problem when triggering `onScrolledToBottom` + `run.debounce`.  Lmk what you think of how it reacts with the most recent commit.

buschtoens:
I noticed two problems with the demo page (in Chrome 64 on CentOS):

screencast_28 02 2018_07_26_23

On my 1920x1080 setup, the table height is taller than what the initially loaded rows would fill. Therefore `lt-infinity` stays in viewport. *Previously* this would cause another `onScrolledToBottom` in order to fetch more rows. This doesn't happen any more. You can force another `onScrolledToBottom` by changing the window size, so that `lt-infinity` exits and enters again. Then everything is working as expected...

screencast_28 02 2018_07_31_14

...except for the fact that continuously scrolling down will prevent `onScrolledToBottom` (or something other down the pipe) from getting called, until the scrolling is stopped. I can scroll down for 30 seconds and longer and see the loading indicator, but the request to mirage is only performed after I stop scrolling.

And the ancient Firefox 52.6 ESR that ships with CentOS doesn't like this one bit. I can't get it to load anything past the first rows. But that's probably just the old Firefox version. Unfortunately I won't be able to test with a current version until this evening.

buschtoens:
I've [rebased your PR](https://github.com/offirgolan/ember-light-table/tree/intersection) and so far the test for [`onScrolledToBottom`](https://github.com/offirgolan/ember-light-table/blob/3aa39a282408646ffb2ba806faa745d99d95b4fe/tests/integration/components/light-table-test.js#L30-L55) runs consitently without problems :tada: :heavy_check_mark: 

https://travis-ci.org/offirgolan/ember-light-table/jobs/368613187#L695

snewcomer:
@buschtoens I brought back the onScrolledToBottom test as well.  I'd be curious to hear what scenarios you can dig up that this upgrade doesn't cover (I know you mentioned perhaps browser compat being an issue from your testing).  I tested on latest versions of Firefox, Safari and Chrome.

addon/components/lt-infinity.js
```diff
@@ -10,62 +8,29 @@ export default Component.extend(InViewportMixin, {
...
-  },
-
-  scheduleScrolledToBottom: observer('rows.[]', 'viewportEntered', function() {
-    if (this.get('viewportEntered')) {
-      /*
-       Continue scheduling onScrolledToBottom until no longer in viewport
-       */
-      this._scheduleScrolledToBottom();
-    }
-  }),
```
buschtoens: This logic is missing and would be a breaking change.

When `onScrolledToBottom` was called and the rows that were added to the body were not enough to push `{{lt-infinity}}` out of the viewport, `{{lt-body}}` kept firing `onScrolledToBottom` repeatedly until the whole viewport was filled.

I don't really like how this has been implemented originally, since `afterRender` wouldn't only be called when rows have actually been added to the table, but every time when *anything* re-renders.

https://github.com/offirgolan/ember-light-table/blob/5dccab6a47644eef0381b30004802aa88958f176/addon/components/lt-infinity.js#L55-L57

Instead of this implicit and brittle detection of table mutations, `{{lt-infinity}}` should  communicate its visibility status to `{{lt-body}}`, which in turn would explicitly observe changes to the [`rows` array](https://github.com/offirgolan/ember-light-table/blob/5dccab6a47644eef0381b30004802aa88958f176/addon/components/lt-body.js#L308) and fire `onScrolledToBottom` `afterRender` as long as `{{lt-infinity}}` is visible.

What do you think of that?

Please don't feel any responsibility to refactor this yourself—you've already done a tremendous job with this PR and I'm super thankful for that. I can also do that, if you don't have the time. :wink:

addon/styles/addon.css
```diff
@@ -63,6 +63,10 @@
           flex: 1 1 auto;
 }
 
+.lt-infinity {
+  min-height: 1px;
+}
```
buschtoens: Do we still need this?

addon/templates/components/lt-body.hbs
```diff
@@ -56,7 +56,7 @@
       </table>
 
       {{#if onScrolledToBottom}}
-        {{lt.infinity rows=rows onScrolledToBottom=onScrolledToBottom scrollBuffer=scrollBuffer}}
+        {{lt.infinity rows=rows onScrolledToBottom=onScrolledToBottom scrollableContent=".lt-scrollable"}}
```
buschtoens: > Lastly, this PR is not working in Safari (using rAF) b/c the scrollBuffer is 500 (> than the height of the scrolling area). If the scrollBuffer was like 100, this would work across browsers.

buschtoens:
With the exception of https://github.com/offirgolan/ember-light-table/pull/548#discussion_r182950407 this now works perfectly in Chrome and even the ancient version of Firefox that ships with CentOS. I'll try to get my hands on a Windows machine with IE11.

addon/templates/components/lt-body.hbs
snewcomer: We could.  I assumed this was for `vertical-collection`, so it would change things up quite a bit in that case.  Looks like a scroll buffer of 100 adds quite a bit more rows to the vertical collection scenario...

addon/styles/addon.css
snewcomer: I think one of the issues is in a scrollable container, there is no height for the IntersectionObserver to get on screen.  I *think*.  Will keep playing around.

addon/components/lt-infinity.js
snewcomer: @buschtoens Ah you are so correct about the regression.  I agree.  How about something like this? 

https://github.com/offirgolan/ember-light-table/pull/548/commits/166f1db468d7b19ef9358febb8d8b87154f0cb20

snewcomer: If this is better, I can add a few scrolling tests as well.

addon/styles/addon.css
snewcomer: I would maybe make this the height of the icon that indicates we are loading more rows.

mkay581:
This fixes a lot of onScrolledToBottom and rendering issues with the latest version of ELT. Can we somehow get this into the next version?

addon/components/lt-infinity.js
mkay581: I wholeheartedly agree with @buschtoens's opinion above. Especially:

> I don't really like how this has been implemented originally, since afterRender wouldn't only be called when rows have actually been added to the table, but every time when anything re-renders.

@snewcomer I like your approach as an initial attempt. as @buschtoens pointed out, `lt-infinity` should probably broadcast when its actually time for `lt-body` to fire its `onScrolledToBottom` event.

snewcomer: THanks @mkay581!  That was already added here - https://github.com/offirgolan/ember-light-table/commit/166f1db468d7b19ef9358febb8d8b87154f0cb20

buschtoens:
I'm really sorry, this is taking so long. I have my last day at my old job tomorrow and—how could it be any different—are totally swamped with work. I'm moving to a new city on Sunday and will try to.get this in on the drive down there. 

I'll give this a quick review tomorrow and push out a pre-release, so you can easily test drive this.

alexander-alvarez:
@buschtoens  don't commit and drive, it can be dangerous

buschtoens:
@alexander-alvarez Haha, fortunately for me, my boyfriend's driving. :smile:

snewcomer:
I tried this out on Chrome, Firefox, and Safari.  It fixes the scrolling to bottom.  However, there are still some improvements we can make.  Perhaps for a later time though if ppl so desire!

It would be great to hear from a few other ppl that this fixes problems/perf for them as well.

PR #550, opened 2018-03-14
Adds cookbook/index

### LINKED ISSUE #443: Add `cookbook.index` route to dummy app
#441 introduces a link to https://offirgolan.github.io/ember-light-table/#/cookbook, but there's no index route yet.

A simple but neat list and maybe a few explanatory paragraphs will do just fine.

bravo-kernel:
Closes #443.

### DIFF
--- a/tests/dummy/app/router.js
+++ b/tests/dummy/app/router.js
@@ -12,9 +12,9 @@ Router.map(function() {
   this.route('scrolling');
 
   this.route('columns', function() {
+    this.route('draggable');
     this.route('grouped');
     this.route('resizable');
-    this.route('draggable');
   });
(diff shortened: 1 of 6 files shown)


tests/dummy/app/router.js
```diff
@@ -23,13 +23,14 @@ Router.map(function() {
   });
 
   this.route('cookbook', function() {
+    this.route('/');
```
buschtoens: I think you mean `this.route('index');` here? If so, you can skip that, since the `index` route is created implicitly.

bravo-kernel: Correct, removed as requested

PR #565, opened 2018-05-04
Upgrade to ember-in-viewport 3.0.3

john-griffin:


buschtoens:
Unfortuately, there's a bit more to it than just bumping the dependency. Please see #548.

PR #572, opened 2018-07-13
Use isArray to check columns, rows type

quaertym:
This fixes table construction for ember data types such as `DS.RecordArray` and `DS.AdapterPopulatedRecordArray`. Otherwise it does not work with ember data’s `store.findAll('user')` result since `instanceof` check returns false.


quaertym:
🍻

quaertym:
Can you cut a new release?

buschtoens:
Not at a computer right now. Will do later in the evening, if no one else does before. :)

quaertym:
Can you release this if possible?

Issue #416, opened 2017-06-01
classNames, cellClassNames cannot be arrays


buschtoens:
https://github.com/offirgolan/ember-light-table/blob/v1.8.6/addon/classes/Column.js#L259-L265
```js
  /**
   * Class names to be applied to all cells of this column
   *
   * @property cellClassNames
   * @type {String | Array}
   */
  cellClassNames: null,
```

https://github.com/offirgolan/ember-light-table/blob/v1.8.6/addon/components/cells/base.js#L15-L25
```js
/**
 * @module Cell Types
 * @class Base Cell
 */

const Cell = Component.extend({
  layout,
  tagName: 'td',
  classNames: ['lt-cell'],
  attributeBindings: ['style'],
  classNameBindings: ['align', 'isSorted', 'column.cellClassNames'],
  // ..
});
```

The same goes for `classNames`.

When passed an array of class names, the array is stringified (concatenated with `,`).

I'll submit a PR when I can get around for it. Merely putting this here for tracking. :)

PR #573, opened 2018-07-17
change the property type of column className and cellClassName to be String only

baljeet03:
This is regarding the below issue.  
https://github.com/offirgolan/ember-light-table/issues/416

That's my first PR in open source community :)

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -248,7 +248,7 @@ export default class Column extends EmberObject.extend({
    * Class names to be applied to header and footer cells of this column
    *
    * @property classNames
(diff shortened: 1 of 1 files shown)


alexander-alvarez:
Hi @baljeet03  I'm so sorry for not following up sooner.
I'm actually not sure about https://github.com/offirgolan/ember-light-table/issues/416, so I'd need to review with @buschtoens 
https://github.com/emberjs/ember.js/blob/v3.3.1/packages/ember-views/lib/mixins/class_names_support.js#L32 is an array, so as is should work.

PR #584, opened 2018-08-22
Improve cells performance

mostafa-sakhiri:
This is related to #12 .
The utils method `cssStyleify` is not scalable when displaying many cells.
Indeed, the method do the following on every cell
- Introspect each object to get the keys
- Instantiate a new array
- Iterate over the `hash` object
- Try to dasherize the key
- `HtmlSafe` the returned string

This leads to a lot of memory management operations (including garbage collection phases) for the JS VM and it's very slow.
We can get a significant performance gain by bypassing `cssSyleify` in this case.

### DIFF
--- a/addon/components/cells/base.js
+++ b/addon/components/cells/base.js
@@ -33,7 +33,7 @@ const Cell = Component.extend({
     }
 
     // For performance reasons, it's more interesting to bypass cssStyleify
-    // which can leads to a lot of garbage collections
+    // since it leads to a lot of garbage collections
     // when displaying many cells
     return columnWidth ? `width: ${htmlSafe(columnWidth)};` : '';
   }),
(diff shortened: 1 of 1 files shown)


mostafa-sakhiri:
To be more clear about the performance gain that we got :
In our context we had an ember-light-table with 300 cells (50 rows / 6 columns), and it was too slow.
When we decreased the number of rows to 25 rows, we gained a 34% of performance. The conclusion was : there is a problem with cells construction.
When implementing the solution above with 50 rows we gained a 33% of performance.

alexander-alvarez:
Yeah, performance has been a problem area that we're always looking to improve on.

This addon has slowly been integrating [`vertical-collection`](https://github.com/html-next/vertical-collection/) that also helps with the problem (you can see the example code here http://offirgolan.github.io/ember-light-table/#/cookbook/occlusion-rendering). Basically with it we only render the rows that a user could see, instead of the whole table, and dynamically updates what's rendered as you scroll

PR #586, opened 2018-09-10
FIX - Incomplete use of htmlSafe() on Cell.style

ghost:
Seeing large numbers of these warnings: `Binding style attributes may introduce cross-site scripting vulnerabilities ...` on Ember 3.3.2.

The empty string being returned when there is no `columnWidth` needed to be wrapped in `htmlSafe()` or changed to `null`. Also I think the `htmlSafe()` call needs to wrap the entire `width: ...` style.

### DIFF
--- a/addon/components/cells/base.js
+++ b/addon/components/cells/base.js
@@ -35,7 +35,7 @@ const Cell = Component.extend({
     // For performance reasons, it's more interesting to bypass cssStyleify
     // since it leads to a lot of garbage collections
     // when displaying many cells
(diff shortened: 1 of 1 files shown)


Techn1x:
I'm getting this issue too. I'm running Ember 3.2.2, super annoying because it clogs up my console XD (1 warning for every cell, since all of my columns have no defined width)

donaldwasserman:
FWIW - I added this + the removal of the `sendAction` deprecations to my fork at `donaldwasserman/ember-light-table#prod`

- YMMV

alexander-alvarez:
Sorry for the delay

PR #590, opened 2018-10-09
replace `sendAction` with modern callable methods

donaldwasserman:
This fixes #587.

CC: @buschtoens

PR #599, opened 2018-11-19
Stop using sendAction, which is deprecated.

plcarmel:


alexander-alvarez:
We merged this already in a 2.x branch https://github.com/offirgolan/ember-light-table/pull/590
You can check it out with `ember-light-table@next`

PR #601, opened 2018-11-19
Avoid breaking line with sort icon in narrow columns

### LINKED ISSUE #600: Sort icon gets mounted when the columns are narrow
I have found that, in tables with a lot of columns or with narrow ones, the sort icon in the header gets mounted by the label like this (you can see it in the last name column):

screenshot from 2018-11-19 11-14-23

What do you think about changing the order of [these lines](https://github.com/offirgolan/ember-light-table/blob/master/addon/templates/components/columns/base.hbs#L4-L9)? It doesn't totally fix the problem but at least the result is prettier than the current solution. Other way is adding the option to allow ellipsis on the headers :smile:.

Gorzas:
Close #600

### DIFF
--- a/addon/templates/components/columns/base.hbs
+++ b/addon/templates/components/columns/base.hbs
@@ -1,12 +1,12 @@
 {{#if column.component}}
   {{component column.component column=column table=table tableActions=tableActions extra=extra sortIcons=sortIcons}}
 {{else}}
-  {{label}}
   {{#if (and sortIcons.iconComponent sortIconProperty)}}
     {{component sortIcons.iconComponent sortIcons=sortIcons sortIconProperty=sortIconProperty}}
   {{else if sortIconProperty}}
     <i class="lt-sort-icon {{get sortIcons sortIconProperty}}"></i>
   {{/if}}
+  {{label}}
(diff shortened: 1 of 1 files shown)


alexander-alvarez:
Hi @Gorzas, thanks for investigating this and proposing a solution. I'm happy to ideate to find something that works.
With the current change does this change the visual at all? Would the sort indicator show up on the left of the text?

Gorzas:
With my solution, the headers are shown like this (I think it's a little bit prettier than the current option):

screenshot from 2018-11-20 08-54-42

My idea is that, when the column is too narrow, the name of the column and the sorting icons can fit together. This way, the developer can hide the text that doesn't fit (but I don't know if the ellipsis is something that should be in the library):

screenshot from 2018-11-20 08-55-49

alexander-alvarez:
Got it. What you show in the first image vs the one in https://github.com/offirgolan/ember-light-table/issues/600 is a good improvement.

Give me some time to run through the other pages manually to make sure everything else is OK, and then I'll merge.

Thanks again!

alexander-alvarez:
Sorry for the delay. LGTM!

PR #650, opened 2018-12-07
Update ember-cli-changelog

Gaurav0:


alexander-alvarez:
Hi. Thanks for submitting a PR. Would you like to explain why is this needed?

Gaurav0:
Mostly it gets rid of several warnings on build.

alexander-alvarez:
Can you be more specific? You're changing a third party dev dependency to one under an organization who's source code you haven't linked.

Given recent npm events, I feel more info is necessary

Gaurav0:
I've done so. Been waiting.

rwwagner90:
Why do we need it published at all here guys? We can use the tag from github no problem

Gaurav0:
Github tags can be deleted and even assigned to a different commit, but if that's ok with @alexander-alvarez , it's ok with me, since it's only a devDependency and it won't hit my app. Opened a PR that way: #651

PR #661, opened 2019-01-04
FIX - Make ember-responsive a dependency not just a dev dependency

richard-viney:
The main table component injects the `media` service, so `ember-responsive` needs to be listed as a dependency, not just a dev dependency.


richard-viney:
Ah thanks, makes sense.

PR #664, opened 2019-01-23
Fixing issue with multiple tables and onScrolledToBottom.

gmurphey:
When multiple tables were rendered to the document, only the first table would scroll correctly. This was due to the  selector referencing a class name and ember-in-viewport only selecting the first instance of that class. This commit prepends a table ID to that selector so that multiple tables with scrolling is supported.

### DIFF
--- a/addon/templates/components/lt-body.hbs
+++ b/addon/templates/components/lt-body.hbs
@@ -60,7 +60,7 @@
           rows=rows
           inViewport=(action "inViewport")
           exitViewport=(action "exitViewport")
-          scrollableContent=".lt-scrollable"}}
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
Thank you!

fran-worley:
When [merging master into 2-x branch](https://github.com/offirgolan/ember-light-table/commit/b173a6ae46a5928a6149e30aeeaa644a3481ca5f#diff-8aded05a725d93d604010e6f504429a4) I had to update our tests that call the `lt-body` directly to ensure that the `tableId` is always set otherwise we were getting errors raised for invalid selectors.

PR #665, opened 2019-01-23
Fixing virtual scrollbar flip flopping. Closes #552.

### LINKED ISSUE #552: When using fixed headers or footers, rows are rendered twice
My team noticed that table rendering was a bit sluggish and on further investigation, realized that each cell component was being rendered twice. I've added some tests that illustrate the issue.

https://github.com/metricly/ember-light-table/commit/11adf8d95fd744ce6cc211e0d1ea8d9fd45c218e

I believe this is caused by the runloop surrounding the virtual scrollbar logic:

https://github.com/offirgolan/ember-light-table/blob/6d71609da875453ff9792f636b88c5c2e7e7f385/addon/components/lt-body.js#L322

Since this waits until `lt-head`, `lt-body`, and `lt-foot` have all been initialized. We've worked our way around this for now by creating `fixedHead` and `fixedFoot` properties on `light-table`, but didn't want to create a PR with those changes since it would introduce a breaking change. If there are any ideas for a better way to address this, I'd be more than happy to put together a PR and move this forward. Thanks for all the great work!

gmurphey:
When using a virtual scrollbar, the table body was rerendered as the virtual scrollbar flag was settling.

### DIFF
--- a/addon/components/lt-body.js
+++ b/addon/components/lt-body.js
@@ -221,6 +221,14 @@ export default Component.extend({
   useVirtualScrollbar: false,
 
   /**
+   * @property virtualScrollbarSettled
+   * @type {Boolean}
+   * @default false
+   * @private
+   */
+  virtualScrollbarSettled: false,
+
+  /**
    * Set this property to scroll to a specific px offset.
    *
    * This only works when `useVirtualScrollbar` is `true`, i.e. when you are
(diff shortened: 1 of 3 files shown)


rwwagner90:
Hi @gmurphey I realize this has been open a long time, but could you possibly rebase?

PR #666, opened 2019-01-23
Pinning ip-regex to fix build issue.

gmurphey:
https://github.com/salesforce/tough-cookie/issues/140


alexander-alvarez:
Thank you!

Issue #444, opened 2017-06-27
Transform native ES6 classes into Ember.Object


buschtoens:
Follow up on https://github.com/offirgolan/ember-light-table/pull/439#issuecomment-311264044.

Currently ELT uses three native ES6 classes intermingled with `Ember.Object` (namely `Column`, `Row`, `Table`), that can easily be transformed into pure `Ember.Object`. This is a breaking change, as users will have to switch from `new Table(columns, rows, options)` to `Table.create(columns, rows, options)`.

buschtoens:
Relevant: https://github.com/emberjs/rfcs/pull/240

PR #675, opened 2019-03-17
Fix deprecation: object.new-constructor #662

### LINKED ISSUE #662: Deprecation: object.new-constructor
I'm getting the following deprecation on ember-cli@3.7.1

> DEPRECATION: using `new` with EmberObject has been deprecated. Please use `create` instead, or consider using native classes without extending from EmberObject. [deprecation id: object.new-constructor]

This message is thrown when extending the [Table class](https://github.com/offirgolan/ember-light-table/blob/d4673d80c4ea622bf2513358f70e28a06e84b45b/addon/classes/Table.js#L30)

fran-worley:
Fixes #662

Now let me start by saying, I'm not sure this is the best solution BUT it does fix the deprecation errors while still supporting the existing API. 

Let me know what you think, any alternative solution is likely to require a discussion on where to take the API.

### DIFF
--- a/addon/classes/Column.js
+++ b/addon/classes/Column.js
@@ -318,7 +318,7 @@ export default class Column extends EmberObject.extend({
   init(...args) {
     this._super(...args);
 
-    const subColumns = emberArray(makeArray(this.get('subColumns')).map((sc) => new Column(sc)));
+    const subColumns = emberArray(makeArray(this.get('subColumns')).map((sc) => Column.create(sc)));
     subColumns.setEach('parent', this);
 
     this.set('subColumns', subColumns);
--- a/addon/classes/Table.js
+++ b/addon/classes/Table.js
@@ -147,6 +147,14 @@ export default class Table extends EmberObject.extend({
    *           `createRow(content, options)`.
    */
   constructor(columns = [], rows = [], options = {}) {
+    // this enables Table.create({}) api which fixes deprectation introduced in ember 3.6
+    if (!isArray(columns) && typeof columns === 'object' && columns !== null) {
+      options = columns;
+      rows = columns.rows || [];
+      columns = columns.columns || [];
+      delete options.rows;
+      delete options.columns;
+    }
     super();
 
     assert('[ember-light-table] columns must be an array if defined', isArray(columns));
@@ -433,7 +441,7 @@ export default class Table extends EmberObject.extend({
    * @return {Column}
    */
   static createColumn(column) {
-    return new Column(column);
+    return Column.create(column);
   }
 
   /**
--- a/tests/unit/classes/table-test.js
+++ b/tests/unit/classes/table-test.js
@@ -14,6 +14,17 @@ module('Unit | Classes | Table', function() {
   });
 
   test('create table - with options', function(assert) {
+    let table = Table.create({ columns: [{}, {}], rows: [{}] });
+
+    assert.ok(table);
+    assert.equal(table.get('rows.length'), 1);
+    assert.equal(table.get('columns.length'), 2);
+
+    assert.ok(table.rows[0] instanceof Row);
+    assert.ok(table.columns[0] instanceof Column);
+  });
+
+  test('new table - with options', function(assert) {
     let table = new Table([{}, {}], [{}]);
 
     assert.ok(table);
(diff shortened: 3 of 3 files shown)


alexander-alvarez:
@buschtoens  any ideas? I know you did some work in this park

fran-worley:
@alexander-alvarez @buschtoens sorry to bump this as I know you guys will be super busy. I'm keen to try and push this over the finish line and get a new beta version out as the deprecation warnings generated by light table are super noisy!

I've got some time at the weekend to take another look at this and revise the approach if anyone can steer me in the right direction?

alexander-alvarez:
I have no knowledge of this particular area -- given that, how can I best help @fran-worley ?

fran-worley:
Hi @alexander-alvarez I didn't mean to sound rude, I just wondered whether there was anything that we could do while waiting for feedback from the others to progress this as the deprecation warnings are very noisy (multiple raised per instance of a table). With each release of ember source, more warnings seem to be coming through and I'd hoped we could keep this addon updated.

If fixing this warning is stuck until we get feedback can we at least release a new beta version to ship the fixes already applied?

richard-viney:
@fran-worley I think the workaround you've added to the constructor will work, I assume that you pair this with replacing `new Table()` with `Table.create()` in the consuming application?

Ideally one of two things would happen:

1. Update everything (docs, tests, etc.) to use `Table.create()` instead of `new Table()`, and replace the existing `constructor`s with `init()` where needed. Existing users would need to update their code for this though.

2. Change the classes in this addon to no longer inherit from `EmberObject`. They'd then need to use decorators, which would limit support to 2.18+ (due to the `ember-decorators-polyfill` only going back that far). This option is more forward-looking, and users would continue to do `new Table()`, so there may be little to no documentation to update. There might be some gotchas in doing this transition, I've only taken a superficial look.

It would certainly be good to sort this out for 2.0 release.

Thoughts?

fran-worley:
@richard-viney they were the two options as I saw them when I drafted this PR.

I opted to start with option 1, as in the absence of input from the addon creators who know where they want to take the project this could be done as a non-breaking change (the non deprecated syntax is simply an option people can choose to make use of). 

In my personal option, I think that option 2 is the best solution as it seems more compliant with the direction Ember is taking but whether the maintainers are comfortable with dropping support to post Ember 2.18 I couldn't say.

Either way, a decision needs to be made and soon as this addon is now falling behind Ember development and as someone who needs a reliable table producing Ember addon that makes me very sad!

richard-viney:
After looking into this a little further I think the best thing to do initially is option #1. Option #2 would be a much more substantial diff and a more breaking change. There may be scope to do an 'Octane upgrade' at some point, but there's no urgency on it.

The main thing to fix is how our public API relies on functionality that was never stabilised, namely using `new` on a class that extends from `EmberObject`. The solution in this PR tries to get around this, but it really would be preferable if we dropped `new Table()` syntax (and also `new Row()` and `new Column()`), potentially replacing the current ES6 constructors with equivalent static creation methods of some kind.

This would be fairly straightforward to do, and the 2.0 release seems like a good point at which to make this change to the public API. Is a maintainer able to comment on whether such a transition would be possible at this stage? If so I'm sure either @fran-worley or myself would be happy to do a PR  for it.

alexander-alvarez:
@fran-worley  @richard-viney
Thanks for the productive discussion. You have identified the all the key points.

We have the 2.x branch that is meant to include any breaking changes we want to take on as a community -- so we can definitely add this into that if we see the ember community at large moving away from the style we adopted years ago

fran-worley:
@richard-viney my only counter to your suggestion to press ahead with option one and do option 2 later is that we're at risk of changing the API twice (i.e. reverting to `Class.create`, only to re-introduce the ES6 new `new Class`.

If dropping support for < Ember 2.18 is an option for the 2.x branch then I suggest that we seriously explore option 2 as it would fix the deprecation warning and keep the ELT API the same. May also have the added bonus of a perf increase (not extending EmberObject), work with Octane and be more in keeping with where the Ember project is going.

richard-viney:
@fran-worley Yes I agree. The good points about option 1 are that it can be done pretty swiftly, and apart from switching from `new()` means no external API changes either. I'm not familiar enough with this library's usage patterns to know if removing the `EmberObject` inheritance from these classes would be a meaningfully visible change to external users. Possibly not, in which case that approach would probably be best.

Ember 2.18 is close to being out of extended support, so dropping support for earlier versions may be fine given we're looking at a major version release anyway.

Do the project maintainers have a preference here?

fran-worley:
@richard-viney I spoke to @alexander-alvarez yesterday and there is no issue with setting the minimum ember version for the 2.x branch at 2.18 so we can explore option 2 which I'll do in a separate PR. Let's see how much breaks by converting these to native js classes with decorators and take it from there.

fran-worley:
Although I have now found this - https://github.com/offirgolan/ember-light-table/issues/444 which suggests the goal is actually to change the ELT api entirely so maybe we just embrace option 1 and change the API for 2.x branch updating the docs as necessary.

richard-viney:
OK - so are we agreed on the way forward here? i.e. option 1 for the v2 release as it's less likely to hit unexpected snags and the upgrade path is simple, followed by an 'Octane upgrade' at some later date (call it v3 for argument's sake).

fran-worley:
@richard-viney yes but I think rather than trying to support both apis we need to change to `Table.create` and ensure all the docs are updated. This would close #444 as well as fixing this Ember deprecation.

richard-viney:
Yep agreed, users will need to update to `Table.create`, and we'll upgrade to supported use of native classes in modern Ember.

fran-worley:
Wonderful - I'll try and find some time next week to revise this PR accordingly!

fran-worley:
Pausing this pending #693 . Hopefully we can then bundle them both into a new breaking beta release.

fran-worley:
Closing in favour of #701

PR #677, opened 2019-03-20
ensure ember-scrollable updates when rows are updated

### LINKED ISSUE #676: Fixed header - Show scroll bar all the time
I had assumed that turning off autoHideScrollbar would result in the scroll bar always being visible.

Turning off autoHide seems to ensure it stays visible, but only becomes visible after scroll has occurred.

I’m raising this as a number of our users have complained of being unable to scroll our tables since we adopted Ember Light Table. On closer inspection, the problem appears to be be with users who prefer to click and drag a scroll bar rather than use gestures or a scroll wheel to do so. 

Even so I believe this is a valid issue - it should be possible to display scroll bars at all times and I would expect disabling autoHideScollbard to control this behaviour.

fran-worley:
Fixes #676 

Essentially this ensures that ember-scrollable is 'resized' when rows are updated - based on the ember-scrollable guidance [here](https://github.com/alphasights/ember-scrollable/blob/master/addon/components/ember-scrollable.js#L448...L470 ) 

Which comes from [this](https://github.com/alphasights/ember-scrollable/issues/85) ember-scrollable issue.

I'm not entirely sure the best way to test this - any ideas @alexander-alvarez ?

### DIFF
--- a/addon/templates/components/lt-body.hbs
+++ b/addon/templates/components/lt-body.hbs
@@ -11,6 +11,7 @@
       autoHide=autoHideScrollbar
       scrollTo=targetScrollOffset
       onScrollY=(action 'onScroll')
+      as |scrollbar|
     }}
       <div id={{concat tableId '_inline_head'}} class="lt-inline lt-head"></div>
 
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
thanks @fran-worley this looks great... I have no idea how to automated test it. Manual QA?

fran-worley:
I’ve tested it locally just by overriding the templates in my app and it’s workig a treat!

PR #674, opened 2019-03-15
Update ember-in-viewport

kevinansfield:
no issue
- bumps `ember-in-viewport` too `3.2.2`
- adds a fix for passing non-existent selectors to `ember-in-viewport` as occurs when there is no fixed header/footer and therefore no `.lt-scrollable` element

This was a necessary upgrade for us because the pinned version of `ember-in-viewport` has a bug in Safari that was causing our other uses of `{{ember-infinity}}` to fail. Upgrading the version of `ember-in-viewport` in our app (and forcing resolution via package.json) ran into the `scrollableArea` bug that is also fixed in this PR.

PR #682, opened 2019-04-03
update dependencies

bekzod:


snewcomer:
@bekzod What do you think about updating the travis config as well?  Perhaps ember-try back to 2.16?

snewcomer:
Updating travis will allow deps to use classes as well.  2.12 fails Ref #674

snewcomer:
@fran-worley Any thoughts on this PR?  I could bump the travis config post merge

fran-worley:
@snewcomer thanks for the ping - we should bump the ember-concurrency dependency as I think we're on 10 or 11 (maybe even 12 now!) Can we double check other dependencies as we may aswell get this right at the point of merging.

In terms of ember try/ travis, we should ensure that we're testing against all 'supported' ember versions and for the master branch I believe this goes back to ember 2.13.

I'm not entirely convinced about bumping ember js quite so high on master branch as the deprecations being thrown up by ember are only being fixed in the 2.x branch.

I suggest we bump the ember version to the point at which the deprecations start getting noisy on master branch and then raise a further PR against 2.x branch to update ember to latest version. Again, updating ember try/ travis to test all 'supported' versions.

fran-worley:
May also be worth updating the readme to point people using ember >3? to the v2.x branch.

bekzod:
sorry for missing, you seem handled everything shout if you need me

fran-worley:
Closing in favour of #693 thanks @bekzod I've taken this into a new PR to finish this off.

Issue #678, opened 2019-03-24
Remove all instances of sendAction


MichalBryxi:
[This PR](https://github.com/offirgolan/ember-light-table/pull/590/files) started removing deprecated `sendAction` calls. 

But I believe there are some instances of sendAction that you probably missed: https://github.com/offirgolan/ember-light-table/blob/2-x/addon/mixins/table-header.js#L181

Maybe they are not intended to be in use, but I'm basically copy-pasting the examples from your guides and I stumbled upon this. I'm using your `@next` branch that points me to `2.0.0-beta.2`

fran-worley:
@MichalBryxi there definitely are still a few that have been missed. Did you want to open up a PR to continue the removal process? I may have some time in the next couple of weeks to pickup/ help with this as I'm keen to get current Ember deprecations sorted on the 2.x branch.

MichalBryxi:
@fran-worley - https://github.com/offirgolan/ember-light-table/pull/686

fran-worley:
@alexander-alvarez I believe this can be closed as the PR was merged and released as part of 2.0.0.beta.3

PR #686, opened 2019-05-03
refactor: Remove sendAction() calls

MichalBryxi:
As per: #678


alexander-alvarez:
Thank you @MichalBryxi

Issue #542, opened 2018-02-15
Occlusion and Responsive


ultish:
Hi,

I just stumbled across the occlusion option and realised that with responsive cells you can't expand them. They will hide when they hit their breakpoint but you can't expand the row to show them again. Looking at the lt-body.hbs template it's obvious why the expansion doesn't work since it's not in the template, but are these 2 mutually exclusive or will there be plans to add it in in the future?

buschtoens:
I personally haven't use the responsive / media features yet, so I got no experience with them. I'll try to look into it. In the meanwhile, feel free to give it a shot yourself. :muscle:

joeypedicini92:
Just want to bring this back up I'm running into a performance problem and occlusion seems to be the solution for me, however I need expandable rows. 

I tried adding the expanded-row into the virtual-collection part of the template and everything seems to work fine I'm not noticing anything weird at all. 

What's the reason (if any) for not allowing expandable rows using occlusion?

Edward-Chen-CN:
> Just want to bring this back up I'm running into a performance problem and occlusion seems to be the solution for me, however I need expandable rows.
> 
> I tried adding the expanded-row into the virtual-collection part of the template and everything seems to work fine I'm not noticing anything weird at all.
> 
> What's the reason (if any) for not allowing expandable rows using occlusion?

Are you sure you are running without any issues? I have tried your solution but I got console error related to the loading message. It seems like occlusion can not find the loading message in DOM to remove it when scrolled to the bottom

PR #687, opened 2019-05-10
Allow expanding rows in occlusion mode 

joeypedicini92:
pretty straightforward, I can't find a reason that this won't work. But ultimately if someone is experiencing problems they can just turn expanded rows off or occlusion off

see https://github.com/offirgolan/ember-light-table/issues/542


Techn1x:
This is the only thing stopping me from using occlusion

Laxix:
I was able to fix this issue using custom row component extending Row when occlusion is enabled in the table.

Added the code to render columns (as in It-row.hbs) inside below code.

<tr class="lt-row inner-row-expandable">
    <table>
      <tbody>

PR #690, opened 2019-05-18
cleanup `lt-column-resizer`

bekzod:


addon/components/lt-column-resizer.js
```diff
@@ -16,7 +16,7 @@ export default Component.extend({
   startX: null,
 
   $column: computed(function() {
-    return $(this.get('element')).parent('th');
```
bekzod: `$(this.get('element'))` is same as `this.$()`

fran-worley:
Thanks for this but on the basis that in 2.x we've removed jQuery all together I'm reluctant to merge this just for 'style' reasons though you are totally correct of course!!

bekzod:
thanks for comment

PR #692, opened 2019-05-20
Replace propertyWillChange/propertyDidChange with notifyPropertyChange

mmadsen2:
Replace propertyWillChange/propertyDidChange with notifyPropertyChange and add polyfill for < Ember 3.1


fran-worley:
Looks good! - I think we're better updating the methods as @mmadsen2 has done here rather than relying on the polyfill to do it for us.  :shipit: 👍

PR #696, opened 2019-05-26
Bump Ember CLI to 3.8 and update other dependencies

fran-worley:
This looks larger than it is as it includes lint fixes for both js and hbs.

We're just upgrading to ember 3.8 (most recent LTS) and bumping the versions of other dependencies including any revisions to the dummy app where required (e.g. ember-mirage).

PR #698, opened 2019-05-26
Drop support for Node 6 as end of life 30 April 2019

fran-worley:


fran-worley:
@alexander-alvarez , @snewcomer are we happy that it's fair to do this on the 2.x branch? I've seen a number of addons starting to do this and it looks like to support node 6 and bump our dependencies we'd have to start pinning versions (see [travis failure](https://travis-ci.org/offirgolan/ember-light-table/jobs/537459308) for #696).

PR #700, opened 2019-05-29
Set minimum supported ember version at 2.18

fran-worley:
Based on the community [survey results](https://emberjs.com/ember-community-survey-2019/) there is still enough of a user base for 2.18 that we'll need to maintain support for a little longer while people continue to upgrade to ember 3.x. 

2.18 is a sensible place to stop as a number of polyfills for new behaviour only go back to 2.18 and continuing to support older versions will make it harder to keep up with the pace of ember development.

Don't merge until #696 (sorry I stupidly chained these branches together locally).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,6 @@ env:
   matrix:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
-    - EMBER_TRY_SCENARIO=ember-lts-2.12
-    - EMBER_TRY_SCENARIO=ember-lts-2.16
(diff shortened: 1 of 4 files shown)


fran-worley:
@alexander-alvarez , @snewcomer in principal do you foresee any issues with this? 

Obviously I'll rebase once we've updated 2-x branch to ember v3.8 which will leave us with just the [one commit](https://github.com/offirgolan/ember-light-table/pull/700/commits/6bcc49a80d5423706570a376fbb56d5aa18c9c36) here.

alexander-alvarez:
2 LTS-es is pretty typical in the dev world :+1:  from me

PR #718, opened 2019-08-20
Replace volatile computed properties

fran-worley:
This fixes Ember 3.9 deprecation for volatile computed properties.

### DIFF
--- a/addon/components/lt-column-resizer.js
+++ b/addon/components/lt-column-resizer.js
@@ -1,5 +1,5 @@
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { get } from '@ember/object';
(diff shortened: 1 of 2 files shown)


fran-worley:
@alexander-alvarez do you have any opinions here? I was originally going to just convert to native getters but the polyfill seems to be quite buggy and it is only natively supported from 3.10. Given that neither of these methods appear to have been documented as part of any public APIs I've just made them methods for now.

Technically it is another breaking change but what do you think?

alexander-alvarez:
@fran-worley Seeing as they're triggered by user events I think the methods will work just fine. I wouldn't worry about it being a breaking change, it's pretty low-level

PR #729, opened 2020-03-25
fix(lt-body): scrollToRow before render

FalkF:
When changing the `scrollToRow` before newly added rows are rendered `targetScrollOffset` can not be calculated:

```js
this.table.addRows(data);
const targetRow = this.table.rows.find(element => element.get('name') === 'target');
set(this, 'scrollToRow', targetRow);
```
This could be worked around by just changing `scrollToRow` in the next runloop, but it would be cool to just make it work.


rwwagner90:
@FalkF Thanks for the PR! We've recently made a lot of significant changes to ember-light-table. Can you please rebase your PR and make sure everything is still working as intended? If this PR is no longer necessary after the new changes, please close it. Thanks!

PR #735, opened 2020-04-06
WIP: Update dummy app to latest ember conventions

fran-worley:
Our dummy app should be an example of how the addon should be used. We need to update it to use angle bracket syntax, native classes and decorators.


maxwondercorn:
Branch was merged with pr #775

PR #736, opened 2020-04-07
Remove unsafe style binding warnings

fran-worley:


addon/components/cells/base.js
```diff
@@ -29,7 +29,7 @@ const Cell = Component.extend({
     let columnWidth = column.get('width');
 
     if (this.get('enableScaffolding') || !column) {
-      return '';
+      return null;
```
snewcomer, line 32: Is `return;` ok/better as well?

fran-worley: probably - though now I look at it I'm tempted to make them all `return '';` so these functions always return a string. Thoughts?

PR #737, opened 2020-04-07
Drop ember-composable-helpers dependency

### LINKED ISSUE #731: Error: You attempted to overwrite the built-in helper “array”.
Using Ember `3.17.1` with ELT `2.0.0-beta.5` we get the following error:

```
Uncaught Error: Assertion Failed: You attempted to overwrite the built-in helper "array" which is not allowed. Please rename the helper.
```

Tracking this down via [this issue](https://stackoverflow.com/questions/60746799/error-you-attempted-to-overwrite-the-built-in-helper-array-where-is-this-com), it originates from using `ember-composable-helpers` on a lower version than `3.x.x`. When I manually force this in the package-lock to be updated, everything works again.

Can you check if ELT runs perfect with the latest version of `ember-composable-helpers`?

fran-worley:
Fixes #731 

Essentially, the newest versions of ember require v3 of ember composable helpers which in turn only supports ember 3.13+. We can't bump our own dependency as we are trying to maintain support back to ember 3.4.

Pending a more elaborate refactor in v3, I'm proposing we lift the compute helper directly from ember-composable-helpers to enable us to drop the dependency entirely.

I'm going to mark this as breaking as dropping the dependency may catch a few people out who are using it elsewhere in their applications.


fran-worley:
I've also done a little test locally, and when `ember-composable-helpers` is installed it appears to favour their helper over ours so this shouldn't cause any issues for people with `ember-composable-helpers` installed.

package.json
```diff
@@ -62,6 +61,7 @@
     "ember-cli-uglify": "^3.0.0",
     "ember-cli-yuidoc": "0.9.1",
     "ember-code-snippet": "^3.0.0",
+    "ember-composable-helpers": "^3.1.1",
```
snewcomer, line 64: Should we just remove this as well?

fran-worley: Probably but until I've finished digging through the dummy app I can't be sure it's not in use somewhere...
I've got a WIP PR for the dummy app where I can clean this up if not needed.

snewcomer: Ah ok.  If it was removed and the PR passed, then we could do without?

fran-worley: @snewcomer I've checked and we use pipe quite a bit in the dummy app so it will need to stay for the timebeing.

PR #741, opened 2020-04-23
Remove debugging line in compute helper

richard-viney:
Removes leftover debugging line (this line was causing exceptions in my app)


fran-worley:
Thanks!!

richard-viney:
Would it be possible to do a a 2.0.0-beta.6 release with this change included?

Issue #758, opened 2021-02-16
Update test blueprints to new format


Gorzas:
I was working lately with `ember-light-table` and version 3.24 of Ember. I've found that, when creating a new cell-type, it also creates an integration test file but it's in the old format. I think we should update it using the lately integration tests of Ember.

I'll open a PR as soon as possible so you can see an example of what I'm talking about.

Gorzas:
If somebody could take a look to the changes, I can actually update the rest of the blueprints :).

Gorzas:
Thanks @rwwagner90 for the merge! I'll close this ticket if no more work is needed 😄

PR #759, opened 2021-02-16
Update cell-type integration tests blueprint

Gorzas:
Relates #758


Gorzas:
I know that I haven't update the component syntax to AngleBracket but I don't if this new syntax was going to work.

rwwagner90 (approved):

PR #793, opened 2022-07-14
Convert dummy app

maxwondercorn:
Still requires ember @classic decorator until addon is converted.

### DIFF
--- a/tests/dummy/app/components/cookbook/client-side-table.js
+++ b/tests/dummy/app/components/cookbook/client-side-table.js
@@ -1,8 +1,9 @@
 // BEGIN-SNIPPET client-side-table
 import classic from 'ember-classic-decorator';
 import BaseTable from '../base-table';
-import { computed, action } from '@ember/object';
+import { action } from '@ember/object';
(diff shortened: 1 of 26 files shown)


tests/dummy/app/components/cookbook/client-side-table.js
```diff
@@ -57,55 +62,56 @@ export default BaseTable.extend({
         valuePath: 'country',
       },
     ];
-  }),
+  }
 
-  fetchRecords: task(function* () {
-    let records = yield this.store.query('user', { page: 1, limit: 100 });
+  @task({ on: 'init' }) *fetchRecords() {
```
rwwagner90, line 67: @maxwondercorn would it be possible to call this in `init` instead of doing `on: 'init'`?

rwwagner90 (approved):
This looks good to me! Just one question

tests/dummy/app/components/cookbook/client-side-table.js
maxwondercorn: Yes, more clear code.  I was just refactoring and didn't catch this.  Fix pushed

PR #802, opened 2022-07-17
Remove responsive observer/no longer required

maxwondercorn:


maxwondercorn:
Correction. Still required 😄

PR #804, opened 2022-07-18
Remove enableSync option, require updating table rows manually

rwwagner90:
This requires manually updating the table rows with things like `table.addRow` when the data changes. This is a breaking change, since before it was possible to keep things in sync with magic two way binding.


maxwondercorn (approved):
Definitely a  good change. Removes some of the magic and makes app code more explicit/understandable.

PR #807, opened 2022-07-30
Create local html-safe helper/remove ember-cli-string-helpers

### LINKED ISSUE #524: Remove unused / barely used dependencies?
I'm planning on using this add-on in a client app (thanks very much for all the work, it's great! ❤️ ) While looking things over, I was struck by the number of dependencies that were going to get included in my app:

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/package.json#L25-L36

The ones that specifically caught my eye were the various helpers. Curious, I dug in to see where all those helpers were being used.

### ember-composable-helpers
No longer used in add-on templates from what I can see, just used in the `dummy` app.

### ember-cli-string-helpers
The `html-safe` helper is used in three places:

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/addon/templates/components/lt-head.hbs#L13-L15

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/addon/templates/components/lt-foot.hbs#L6-L8

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/addon/templates/components/lt-body.hbs#L21-L23

### ember-truth-helpers
The `or` helper is used three times in one place:

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/addon/templates/components/lt-body.hbs#L2-L4

The `and` helper is used once:

https://github.com/offirgolan/ember-light-table/blob/0d99b0f4655d6a02d0ee1d3c1801db822922269d/addon/templates/components/columns/base.hbs#L5

### Should we remove these dependencies that bleed into consuming apps?
I've steered clear of using these types of helpers in client apps, as I prefer cleaner templates. I realize that's not something everyone agrees with, but until tree-shaking hits, I'd prefer to keep my asset sizes down as much as I can.

Do we want to remove these dependencies to slightly reduce asset sizes for consuming apps? Doing so could be a breaking-change if other developers have begun to rely on the helpers without specifically adding them to their own app dependencies ...

If we do want to remove the dependencies, it looks like we might want to replicate the `html-safe` helper (as it's being used in some loops that might be easiest to leave alone). The other calls could easily move into computed properties though ...

Interested in your thoughts! 😀

maxwondercorn:
Removes `ember-cli-string-helpers` addon dependency.  String helpers now only needed for the `classify` helper in the dummy app.

Closes #524


ember-cli-build.js
```diff
@@ -18,6 +18,10 @@ module.exports = function (defaults) {
     'ember-cli-babel': {
       includePolyfill: true,
     },
+
+    'ember-cli-string-helpers': {
+      only: ['classify'],
```
rwwagner90, line 23: Should we not just create a `classify` helper in the dummy app and remove the addon completely? https://api.emberjs.com/ember/release/classes/String/methods/loc?anchor=classify

maxwondercorn: That makes too much sense :smile: Done

PR #809, opened 2022-07-30
Convert positional params to named params

maxwondercorn:


rwwagner90:
This will be a major breaking change, but I do think it is necessary.

maxwondercorn:
Yes, we need this to complete internal conversion to angle brackets. Hopefully for most apps `table` would be the only change.  `row`, `column` and maybe others if they are extending components.

I already have an almost complete conversion to native classes once this is merged

rwwagner90 (approved):

PR #810, opened 2022-07-31
Remove ember-responsive install from blueprint

maxwondercorn:


rwwagner90:
@maxwondercorn are we sure we don't need this? I think the idea behind installing it is it will create the breakpoints.js file in the consuming app and set all that up.

maxwondercorn:
@rwwagner90 I did forget about the breakpoints file.  I was only looking at ember-responsive moving from devDependencies to addon dependencies.  Does it really make sense to have it in both?

rwwagner90:
@maxwondercorn I think it should be in dependencies, since we're requiring it. Perhaps we could use `afterInstall` to run `ember g ember-responsive` or something like that?

PR #811, opened 2022-08-01
Fix client-side and paginated examples

maxwondercorn:


tests/dummy/app/components/cookbook/paginated-table.js
```diff
@@ -54,7 +54,8 @@ export default class PaginatedTable extends BaseTable {
     }
 
     this.page = page;
-    this.model.clear();
+    const currentRecordsArray = this.table.rows.toArray();
+    this.table.removeRows(currentRecordsArray);
```
rwwagner90, line 58: Would `this.model.toArray()` work as well? Might be closer to the `this.model.clear` we had before.

rwwagner90 (approved):
This looks good to me, just wanted to clarify if we could still use `this.model` in setting rows and clearing? Might be closer to the `model.clear` we had before. If not, no worries.

PR #812, opened 2022-08-01
Document changes to sorting icons and icon component

maxwondercorn:
Document changes to icon parameters and required icon component

### DIFF
--- a/addon/mixins/table-header.js
+++ b/addon/mixins/table-header.js
@@ -78,14 +78,13 @@ export default Mixin.create({
   resizeOnDrag: false,
 
   /**
-   * CSS classes to be applied to an `<i class="lt-sort-icon"></i>` tag that is
-   * inserted into the column's `<th>` element when the column is sortable but
-   * not yet sorted.
+   * Name of the icon that is inserted into the column's `<th>`
+   * element when the column is sortable but not yet sorted.
    *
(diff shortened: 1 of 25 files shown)


rwwagner90:
@maxwondercorn I'm not sure if we need to make these updates? We're using the same API as before, we're just now passing components instead of classes. If someone wanted to pass just the classes, they could and they could include their own CSS to make those classes add icons.

maxwondercorn:
@rwwagner90 I see what you're saying about being able still use classes. A case of overthinking.  I think we should be less  font awesome specific. Something more generic such as "class names to use with custom CSS, web fonts, or ids to use with a custom icon component" 

This is related to the incomplete upgrade guide.  I call out using the included font awesome icons require implementing the `fa-icon-wrapper` or similar component,  need to add your own icon packs to use other icons, etc.

I'll get back to this

rwwagner90:
@maxwondercorn yes, I agree. Perhaps we should move all the FA deps to devDeps and update the docs to be more generic, but also show a new FA example.

maxwondercorn:
> @maxwondercorn yes, I agree. Perhaps we should move all the FA deps to devDeps and update the docs to be more generic, but also show a new FA example.

I agree as that looks like the original implementation.

PR #814, opened 2022-08-08
Addon native classes

maxwondercorn:
Decorators (`tagName`, `classNamesBindings`, etc) remain because outer HTML breaks demo app/tests due to the extra `divs`. Using `@tagName('')`  to remove them causes other errors.

I believe everything needs to be converted to glimmer to finish the cleanup.

`addObserver`,` removeObserver` are used instead of the `@observes` decorator because it's not glimmer compatible


addon/components/lt-infinity.hbs
```diff
@@ -1 +1,3 @@
-{{yield}}
+<div class="lt-infinity">
```
rwwagner90, line 1: Since we are adding this div here, shouldn't we make the JS tagless?

addon/components/lt-head.hbs
```diff
@@ -48,11 +48,11 @@
...
                 sortIcons=this.sortIcons
                 resizeOnDrag=this.resizeOnDrag
-                click=(action "onColumnClick" column)
-                doubleClick=(action "onColumnDoubleClick" column)
-                onColumnResized=(action "onColumnResized")
-                onColumnDrag=(action "onColumnDrag")
-                onColumnDrop=(action "onColumnDrop")
+                click=(fn this.columnClick column)
+                doubleClick=(fn this.columnDoubleClick column)
+                columnResized=this.columnResized
```
rwwagner90, line 53: Is there a reason we wanted to rename these instead of keeping `on` in the name?

addon/components/lt-infinity.hbs
maxwondercorn: This is one of the situations where tagless breaks the addon.

If it's tagless the enter viewport method isn't triggered when scrolling to the bottom.  The table doesn't load any rows.

The ember id/view class can't be attached so it's not  recognized moving into the viewport.  

Tagless:
image

Not tagless:
image

Not sure there is a work around until it's converted to glimmer...

addon/components/lt-head.hbs
maxwondercorn: I had to change the name several places because the internal method had the same name as the passed parameter. They were both `this.method` when converted to a native class generating infinite loop errors. 

 On later updates I started prefixing these rather than dropping the `on`.  I missed this one when cleaning up.  I will fix this

maxwondercorn: Fixed names

vstefanovic97:
@maxwondercorn , @rwwagner90. Do you have a plan or currently work in progress to convert this to glimmer?
I would like to help in the effort to update this addon.
Not sure if you want to convert everything in one go or do you think it is a better idea to just create incremental PRs upgrading parts until we have covered everything?

addon/components/lt-infinity.hbs
rwwagner90: Converting to glimmer I don't think would make a difference? We have to refactor away from this behavior either way. Suppose we can wait and do it later though.

rwwagner90:
@vstefanovic97 we would love the help! I think either way is fine.

addon/components/lt-head.hbs
rwwagner90: @maxwondercorn right, but the duplicate named method should be deleted. They were just wrappers to call the actions, so they are not needed.

addon/components/lt-column-resizer.js
```diff
@@ -1,44 +1,46 @@
...
-
-  isResizing: false,
-  startWidth: null,
-  startX: null,
+@classic
+@classNameBindings(':lt-column-resizer', 'isResizing')
+export default class LtColumnResizer extends Component {
+  column = null;
+  resizeOnDrag = false;
+  isResizing = false;
```
rwwagner90, line 13: Guessing this should be tracked? Anything we want to update ever should be tracked.

addon/components/lt-column-resizer.js
```diff
@@ -69,9 +71,9 @@ export default Component.extend({
 
       let topLevel = closest(this.element, TOP_LEVEL_CLASS);
       topLevel.classList.remove('is-resizing');
-      this.onColumnResized(width);
+      this.onColumnResized && this.onColumnResized(width);
```
rwwagner90, line 74:
```suggestion
      this.onColumnResized?.(width);
```

RobbieTheWagner (changes requested):
I think we need to take another look at what should be `@tracked` and make sure all the things are updating as we expect. Alternatively, we could put back some of the old computeds.

addon/components/cells/base.js
```diff
@@ -13,82 +19,86 @@ import { htmlSafe } from '@ember/template';
...
     // For performance reasons, it's more interesting to bypass cssStyleify
     // since it leads to a lot of garbage collections
     // when displaying many cells
     return columnWidth ? htmlSafe(`width: ${columnWidth};`) : null;
-  }),
+  }
 
-  align: computed('column.align', function () {
+  get align() {
     return `align-${this.column.align}`;
```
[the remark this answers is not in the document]
maxwondercorn: Yes, got bit when trying to remove some computeds.  I did test this one and it works when the column alignment is changed

addon/components/draggable-column.js
```diff
@@ -68,27 +70,27 @@ export default Mixin.create({
 
     sourceColumn = column;
     this.set('isDragging', true);
-    this.onColumnDrag(sourceColumn, ...arguments);
+    this.onColumnDrag && this.onColumnDrag(sourceColumn, ...arguments);
```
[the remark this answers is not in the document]
maxwondercorn: I forget you could do that with optional chaining.  There are similar examples that I will go back and change

addon/components/draggable-column.js
```diff
@@ -149,17 +151,13 @@ export default Mixin.create({
 
       this.setProperties({ isDragTarget: false, isDragging: false });
 
-      this.onColumnDrop(sourceColumn, true, ...arguments);
+      this.onColumnDrop && this.onColumnDrop(sourceColumn, true, ...arguments);
```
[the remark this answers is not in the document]
maxwondercorn: Never mind I see you picked them all up

addon/components/draggable-column.js
```diff
@@ -53,10 +55,10 @@ export default Mixin.create({
       column.get('droppable') &&
       column.get('parent') === sourceColumn.get('parent')
     );
-  },
+  }
 
   dragStart(e) {
-    this._super(...arguments);
+    // super.dragStart(...arguments);
```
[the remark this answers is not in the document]
maxwondercorn: IIRC when I used super.dragStart I got function not found.  I believe this calling the browser drag and drop api.  I will go back and verify this

addon/components/lt-column-resizer.js
maxwondercorn: Complete

addon/components/draggable-column.js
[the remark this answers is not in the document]
maxwondercorn: Just reverified `super.drag*` are not functions.  Will commented code

addon/components/cells/base.js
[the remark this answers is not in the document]
maxwondercorn: Added alignment check tests on column base

tests/integration/components/light-table/columns/base-test.js
```diff
@@ -11,7 +11,46 @@ module('Integration | Component | Columns | base', function (hooks) {
     // Set any properties with this.set('myProperty', 'value');
     // Handle any actions with this.on('myAction', function(val) { ... });"
     this.set('column', Column.create());
-    await render(hbs`{{light-table/columns/base column=this.column}}`);
+    await render(hbs`<LightTable::Columns::Base @column={{this.column}} />`);
     assert.dom('*').hasText('');
   });
+
+  test('column alignment', async function (assert) {
```
rwwagner90, line 18: @maxwondercorn this doesn't test anything because we're rendering a new component each time. What I was concerned about was the alignment changing on the column.

We should render one time, then change the alignment on the column and make sure the correct alignment class is applied.

addon/components/columns/base.hbs
```diff
@@ -26,6 +26,6 @@
     @table={{this.table}}
     @resizeOnDrag={{this.resizeOnDrag}}
     @isResizing={{mut this.isResizing}}
-    @onColumnResized={{action this.onColumnResized this.column}}
+    @onColumnResized={{fn this.columnResized this.column}}
```
rwwagner90, line 29: Can we put these all back to their old names? This should be `onColumnResized`

addon/components/lt-foot.hbs
```diff
@@ -1,33 +1,35 @@
...
+            {{#each this.columns as |column|}}
+              {{component
+                (concat "light-table/columns/" column.type)
+                column=column
+                table=this.table
+                tableActions=this.tableActions
+                extra=this.extra
+                sortIcons=this.sortIcons
+                resizeOnDrag=this.resizeOnDrag
+                click=(fn this.thOnColumnClick column)
```
rwwagner90, line 23: What are these `th` prefixed methods? I think ideally we keep all names the same, so I would remove `th` from the names.

addon/components/lt-infinity.hbs
maxwondercorn: agree

tests/integration/components/light-table/columns/base-test.js
maxwondercorn: @rwwagner90 I wasn't thinking of these as dynamic properties probably because I don't use them that way in my application.  Since they were computed properties they would have been dynamic and they still need to be.... 🤦 

I already changed `isSorted` to computed because it was not working.  Will update the other's and tests

addon/components/columns/base.hbs
maxwondercorn: The names were changed because it generates an infinite loop error. Ember  recursively calls the action because it can't distinguish between the function and passed parameter.

```js
 @action
 onColumnResized(/* column, width */) {
   this.onColumnResized?.(...arguments);
 }
```

My fix was to rename the internal functions and keep the parameter names the same.

```js
  @action
  thOnColumnResized(/* column, width */) {
    this.onColumnResized?.(...arguments);
  }
```

This can be fixed when converted to  glimmer: `this.args.onColumnResized?.()`. In the meantime I can change the names to something else and/or put TODO comments in the code to rename them when converted to glimmer.

addon/components/lt-foot.hbs
maxwondercorn: See comments above on the why.  When I renamed them I prefixed the names with something representing the component.  In this case table-header.

addon/components/columns/base.hbs
rwwagner90: There should be no wrapping function. Wherever there was an `action` that just called a passed function, we can just remove the action entirely.

addon/components/columns/base.js
```diff
@@ -16,80 +21,101 @@ import cssStyleify from 'ember-light-table/utils/css-styleify';
...
+  get isSortable() {
+    return this.column.sortable;
+  }
+
+  get isSorted() {
+    return this.column.sorted;
+  }
+
+  get isHideable() {
+    return this.column.hideable;
```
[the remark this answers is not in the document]
maxwondercorn: I added `@computed` to the property getters that previously were computed `readOnly`. These are `Column` class properties and can not be `@tracked` until it's converted to a native class. 

I couldn't figure out how to test these but was able to modify a dummy app example to test removing columns, changing alignment, etc.  Using these programmatically isn't documented and probably should have dummy app examples.  I'll create an issue as a reminder

tests/integration/components/light-table/columns/base-test.js
maxwondercorn: Fixed by adding @compute to `column.align`

addon/components/columns/base.hbs
maxwondercorn: Dug into the wrapping functions again.  The functions are passing optional actions to other components.  

In most cases, the target component uses the `{{fn}}` helper with the passed function.  So a valid function is required.

I found a related SO [question](https://stackoverflow.com/questions/58847532/how-do-i-make-an-optional-action-for-an-octane-component) for actions but didn't think these were good solutions in this case.  

I passed the functions directly  were I could and cleaned up the remaining wrapped function names.  While the names don't match the original names they're better than what I used before.

While there may be a better solution here I did not want to drag out this PR, if it could be solved later

maxwondercorn:
> I think we need to take another look at what should be `@tracked` and make sure all the things are updating as we expect. Alternatively, we could put back some of the old computeds.

Added computed back.  Most properties are on the class which still need to be converted to native classes

RobbieTheWagner:
@maxwondercorn I totally forgot about this PR, sorry about that! Where did we leave off?

Issue #816, opened 2022-09-22
Remove font awesome dependencies 


fran-worley:
Based on a quick search, these appear to be used in the test app only so 'should' safely be able to move to devDependencies. Might be worth marking as a breaking change incase anyone is inadvertently relying on the light table dependency in their app.

Will add PR shortly.

PR #817, opened 2022-09-22
move font awesome dependency to dev and bump to latest version

fran-worley:
See #816


rwwagner90:
@fran-worley I think the docs might mention using FaIcon for icons. We may need to change the docs to say you'll need to install font-awesome in your app to use it, not sure. Can you take a look and see if any docs need updates?

PR #826, opened 2023-01-03
Bump vertical-collection to v4

IgnaceMaes:
Currently version 3.1.0 of `@html-next/vertical-collection` is used. This has the following known issue: https://github.com/html-next/vertical-collection/issues/397. This is fixed in the v4 release.

According to the [changelog](https://github.com/html-next/vertical-collection/blob/master/CHANGELOG.md) this shouldn't introduce breaking changes for this addon as far as I can tell.


IgnaceMaes:
FYI @rwwagner90, the second one 😄

rwwagner90:
I had to do some other updates to get CI passing, so I did this as part of another PR

PR #838, opened 2024-08-01
Fix: Remove ember-classic-decorator package and usage

gokolo:
The `ember-classic-decorator` package is not compatible with Ember 5.x.
Remove it from the `ember-light-table` to make it work on Ember 5 projects.


RobbieTheWagner:
@gokolo I don't think we can just remove the classic decorator like this. I think we have to instead refactor all the code to use modern Ember things.

RobbieTheWagner:
This PR was opened awhile ago attempting to update things, if you wanted to try to take it further https://github.com/adopted-ember-addons/ember-light-table/pull/814

PR #847, opened 2026-05-01
fix: add setter for cell `value` CP for Ember 4 strict-mode

dmcnamara-eng:
## What

Add an explicit setter to the cell base's `value` computed property so Ember 4 strict-mode doesn't reject writes that propagate up from child components.

## Why

The base cell template (`addon/components/cells/base.hbs`) passes `value=this.value` into a child component (an editable input or a custom `cellComponent`). When the child writes back through the classic two-way binding, Ember 4 strict-mode now throws:

> Cannot override the computed property \`value\` on light-table/cells/base

Adding a setter makes this allowed; the cached value is invalidated again whenever `rawValue` or `column.format` change, so read semantics are preserved.

## Test plan

- [x] Existing addon tests pass locally (no behavioural change for read paths).
- [x] In a downstream consumer (`phorest/phorest-desktop-ember`) running Ember 4.12, the previously-crashing acceptance test \`marketing | added manually audience: it allows to select clients and updates counts\` now passes.

### DIFF
--- a/addon/components/cells/base.hbs
+++ b/addon/components/cells/base.hbs
@@ -1,4 +1,50 @@
-{{#if this.column.cellComponent}}
+{{!--
+
+  XXX @Phorest Changes:
+
+  1. we are passing rowIndex to the cell component
+  2. we have custom "editable" components
+
+--}}
+
+{{! template-lint-disable no-unnecessary-component-helper}}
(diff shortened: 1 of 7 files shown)


dmcnamara-eng:
Misfired — opening against phorest/ember-light-table instead.

